PR #5, opened 2025-06-22
Update OAuth Component group buttons to be all white instead of black.

Waazify:
## Summary by CodeRabbit

- **Style**
  - Introduced consistent white-themed styling for OAuth buttons, including smooth hover effects.
  - Enhanced visibility of GitHub icons by applying darker, more prominent styles.
  - Applied white backgrounds to OTP digit input boxes for improved clarity.
- **Chores**
  - Added a redirect rule to route `/hebo` requests to the homepage.

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -6,6 +6,11 @@ const nextConfig: NextConfig = {
   skipTrailingSlashRedirect: true,
   async rewrites() {
     return [
+      // Handle the /hebo redirect
+      {
+        source: "/hebo",
+        destination: "/",
+      },
       {
         source: "/ingest/static/:path*",
         destination: "https://eu-assets.i.posthog.com/static/:path*",
--- a/next.config.ts
+++ b/next.config.ts
@@ -6,11 +6,6 @@ const nextConfig: NextConfig = {
   skipTrailingSlashRedirect: true,
   async rewrites() {
     return [
-      // Handle the /hebo redirect
-      {
-        source: "/hebo",
-        destination: "/",
-      },
       {
         source: "/ingest/static/:path*",
         destination: "https://eu-assets.i.posthog.com/static/:path*",
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -196,24 +196,7 @@ input.stack-scope {
   margin: 0.5rem 0 !important;
 }
 
-/* OAuth Button Group Styles - Make all OAuth buttons white */
-.stack-scope [data-stack-oauth-button-group] button,
-.stack-scope [data-stack-oauth-button-group] a,
-.stack-scope [data-stack-oauth-button-group] div[role="button"] {
-  background-color: #ffffff !important;
-  color: #000000 !important;
-  border: 1px solid #e5e7eb !important;
-  transition: all 0.2s ease-in-out !important;
-}
-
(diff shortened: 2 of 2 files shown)


src/app/globals.css
```diff
@@ -196,6 +196,115 @@ input.stack-scope {
...
+}
+
+.stack-scope [data-stack-oauth-button-group] button:hover,
+.stack-scope [data-stack-oauth-button-group] a:hover,
+.stack-scope [data-stack-oauth-button-group] div[role="button"]:hover {
+  background-color: #f9fafb !important;
+  border-color: #d1d5db !important;
+}
+
+/* Alternative selectors for OAuth buttons in case the above don't work */
```
turisanapo: Are these alternative methods for the same solutions? If so, I suggest keeping only the one that works

src/app/globals.css
```diff
@@ -196,6 +196,115 @@ input.stack-scope {
...
+.stack-scope [data-stack-oauth-button-group] svg path {
+  fill: #000000 !important;
+}
+
+/* Target the specific GitHub logo SVG structure */
+.stack-scope svg[viewBox="0 0 496 512"] path {
+  fill: #000000 !important;
+}
+
+/* Alternative approach - target any SVG path with white fill */
```
turisanapo: Are these alternative methods for the same solutions? If so, I suggest keeping only the one that works

turisanapo (changes requested):

next.config.ts
```diff
@@ -6,6 +6,11 @@ const nextConfig: NextConfig = {
   skipTrailingSlashRedirect: true,
   async rewrites() {
     return [
+      // Handle the /hebo redirect
```
turisanapo: Why this change?

src/app/globals.css
```diff
@@ -196,24 +196,7 @@ input.stack-scope {
...
-
-.stack-scope [data-stack-oauth-button-group] button:hover,
-.stack-scope [data-stack-oauth-button-group] a:hover,
-.stack-scope [data-stack-oauth-button-group] div[role="button"]:hover {
-  background-color: #f9fafb !important;
-  border-color: #d1d5db !important;
-}
-
-/* Alternative selectors for OAuth buttons in case the above don't work */
+/* Selectors for OAuth buttons in case the above don't work */
```
turisanapo: Is this comment accurate?

Waazify: Yep. I removed the styling methods that didn't work and just kept the selectors and styles that worked.

next.config.ts
Waazify: For some reason (maybe because I was signed in) it kept redirecting me to /hebo on the site and that page doesnt exist so I added this as a fallback incase it happens to real users.

src/app/globals.css
Waazify: I just realised you were talking about the comment. Let me update that.

next.config.ts
turisanapo: Where does this happen? You shouldn't be redirected to /hebo

Waazify: I did some investigation. It turns out my browser had an issue where it keeps redirecting to /hebo. I fixed it and its working fine and no happening anymore and theres no need for this. I will remove it.

turisanapo (approved):

PR #9, opened 2025-07-06
HEB-23 HEB-29 Implement New Agent Page with Model Selection and SVG Icon; Update Ba…

Waazify:
…ckground Gradient Color

This PR is related to tasks HEB-23, HEB-29.

## Summary by CodeRabbit

* **New Features**
  * Introduced a page for creating new agents, allowing users to select from supported AI models and view free token allocations.
  * Added a dropdown menu to choose an AI model, with details on each model’s free tokens per month.
  * Implemented an error popup component to notify users of loading issues with model data.

* **Style**
  * Updated the background gradient for a refreshed visual appearance.

* **Chores**
  * Added configuration and type definitions to support model selection and ensure code quality.


heiwen:
Input:

1/ Please follow agreed architecture of (i) pages are static (no "use client") (ii) components are "use client" (iii) components (buttons / input fields / ...) in separate library project, independent of application logic (iv) using state manager (v) Skeletons for loading

2/ For parsing static data from a built-in JSON file, there shouldn't be a need for an error handling / dialog nor a loading state. The data should be loaded in a separate service and reused across components.

3/ For redirects in NextJs use https://nextjs.org/docs/pages/api-reference/components/link instead of window.location.href 

4/ Tailwind [px] values should only be used in rare occasions, default to named sizes provided by helper classes

5/ If in doubt what the requirements are (e.g. for free token amount, error state of components), please proactively reach out and ask instead of putting random guesses into a PR

6/ The PR should be "usable" in itself, meaning shouldn't rely on "TODO: implement agent creation logic"

Waazify:
> Input:
> 
> 1/ Please follow agreed architecture of (i) pages are static (no "use client") (ii) components are "use client" (iii) components (buttons / input fields / ...) in separate library project, independent of application logic (iv) using state manager (v) Skeletons for loading
> 
> 2/ For parsing static data from a built-in JSON file, there shouldn't be a need for an error handling / dialog nor a loading state. The data should be loaded in a separate service and reused across components.
> 
> 3/ For redirects in NextJs use "" instead of window.location.href
> 
> 4/ Tailwind [px] values should only be used in rare occasions, default to named sizes provided by helper classes
> 
> 5/ If in doubt what the requirements are (e.g. for free token amount, error state of components), please proactively reach out and ask instead of putting random guesses into a PR
> 
> 6/ The PR should be "usable" in itself, meaning shouldn't rely on "TODO: implement agent creation logic"


> (i) pages are static (no "use client") (ii) components are "use client" (iii) components (buttons / input fields / ...) in separate library project, independent of application logic (iv) using state manager (v) Skeletons for loading

1/ @heiwen Can you help me determine what information should be displayed for the free tokens? 

2/ Also I just wanted to clarify — this PR focuses on UI implementation and the basic model list implementation. The state manager is being handled in a separate task as planned, and backend integration will happen once it is done I added placeholder agent creation logic for now. Once backend is ready, it’ll be properly hooked up.

I also believe the PR is usable as the UI is fully there and working. If you think usable means something else in this scenario could you help me by elaborating more about what it should mean.

If You would like me to introduce the state manager from now I can do it but we have a seperate ticket for that but I believe it will just slow us down and make the PR less organised. How would you prefer I proceed?

3/ In our current use case where there is some logic to be executed after the `Create` button is pressed then the best way to navigate would be to using `useRouter().push("/")`, because the `<Link>` is used for visible navigation elements, not for redirects after logic.

apps/hebo-cloud/src/app/globals.css
```diff
@@ -5,7 +5,7 @@
 
 @theme {
   /* === Colours === */
-  --color-background-default : linear-gradient(0deg,#F8F9D2 0%,#F5F7FA 100%);
```
turisanapo, line 8: Please revert this, the previous one is correct.

PR #28, opened 2025-07-25
Incremental UI improvements

heiwen:
1/ Small UI fixes
2/ Prettier & Component Standardization
3/ Dummy Auth for local dev (using valtio)
4/ Reduce bundle size for auth
5/ Basic ShadCN sidebar
6/ ShadCN-based UserButton
7/ Shell for Settings page


apps/app/src/lib/utils.ts
```diff
@@ -1 +1 @@
-export const isMockMode = !process.env.NEXT_PUBLIC_API_URL
+export const isDev = process.env.NODE_ENV === "development";
```
turisanapo, line 1: What is injecting the `development` into the `NODE_ENV` variable? Does it make sense to document it in the readme if it is expected to be set manually by the developer?

packages/ui/components.json
```diff
@@ -5,7 +5,7 @@
   "tsx": true,
   "tailwind": {
     "config": "",
-    "css": "src/styles/tailwind.css",
+    "css": "../../apps/app/src/styles/tailwind.css",
```
turisanapo, line 8: The `app` depends on `ui/components` for shared components, which are meant to be reused across multiple services.

However, `ui/components` depends on `app` for the CSS file, creating a circular dependency.

Is this circular setup intentional? Also, how should we handle CSS for future services that will also use `ui/components`?

heiwen: This is just used by the shadcn cli to add base styles when adding new components. No runtime dependency.

apps/app/src/lib/utils.ts
heiwen: NextJS handles this: "If the environment variable NODE_ENV is unassigned, Next.js automatically assigns development when running the next dev command, or production for all other commands."

https://nextjs.org/docs/pages/guides/environment-variables

turisanapo: Ok thanks for clarifying. How would be the logic to differentiate running dev connected to app, or without, now?

Wouldn't both have `development` as value?

packages/ui/components.json
turisanapo: Would make it sense to move the `tailwind.css` file into the `ui/components` project for clarity?

Issue #3, opened 2025-05-26
Integrate PostHog Analytics (Next.js + TypeScript Lambdas)


turisanapo:
### Overview

Add **PostHog** to Hebo Cloud so we can track key product metrics, funnels and feature adoption. Integration must cover **client-side** tracking with `posthog-js` and **server-side** capture inside our existing Next.js `app`/API routes (executing on AWS Lambda via SST/OpenNext).

### Tasks

* **Install libraries**
  Add `posthog-js` for the browser and `posthog-node` for server handlers using `pnpm add posthog-js posthog-node`. ([posthog.com](https://posthog.com/docs/libraries/next-js), [posthog.com](https://posthog.com/docs/feature-flags))

* **Configure environment variables**
  Store `NEXT_PUBLIC_POSTHOG_PROJECT_API_KEY`, `NEXT_PUBLIC_POSTHOG_HOST` (proxy URL), and `POSTHOG_SERVER_KEY` in SST stage configs so Lambda functions receive them at build time. ([posthog.com](https://posthog.com/docs/self-host/configure/environment-variables), [SST](https://sst.dev/docs/start/aws/nextjs/))

* **Client bootstrap**
  Initialise PostHog in `app/providers.tsx` (or `_app.tsx` if using pages) with `autocapture: false` for performance. Call `posthog.identify(userId, {email})` immediately after successful sign-in. ([posthog.com](https://posthog.com/docs/libraries/next-js), [posthog.com](https://posthog.com/docs/product-analytics/identify))

* **Server-side capture**
  Create a lightweight wrapper in `lib/analytics.ts` that imports `posthog-node`, instantiates it once per Lambda cold-start, and exposes helpers such as `capture(event, properties)`. Use this inside API routes or server actions (e.g., when an API key is generated). ([posthog.com](https://posthog.com/tutorials/nextjs-app-directory-analytics), [posthog.com](https://posthog.com/docs/feature-flags))

* **Event schema** (adopt the *object verb* naming style)

  * user signed up – account created
  * user logged in – successful session
  * api key created – first key issued
  * dashboard viewed – `/dashboard` loaded
  * settings updated – user profile saved
    ([posthog.com](https://posthog.com/docs/feature-flags))

* **Documentation**
  Add `docs/analytics.md` describing event naming, how to add new events, and how to run locally with `POSTHOG_DISABLED=1`.

### Acceptance Criteria

[...]

PR #44, opened 2025-08-15
Heb 68 repo enhancements #3

heiwen:
## Summary by CodeRabbit

- New Features
  - Consistent max-width for agent pages; agent selector now preserves/displays active agent, improves default navigation, and shows agent list with active indicator.

- Style
  - MDX paragraph typography reduced.
  - Not Found page now uses a semantic landmark.
  - Inputs use slightly smaller text; logo text truncates when space is tight.
  - Sidebar padding now transitions when collapsed.

- Chores
  - Added project lint scripts; expanded ignore lists and TypeScript excludes; updated ESLint and build env; updated pre-commit globs.


apps/app/src/app/(shell)/agent/[slug]/page.mdx
```diff
@@ -4,15 +4,14 @@ import { GenerateApiKey } from "~/components/ui/GenerateApiKey";
 
 <Logo className="hidden md:flex" />
 
-**Hi <UserName />!**
+<p className="text-base font-semibold">Hi <UserName />!</p>
```
Comment (turisanapo) [inline apps/app/src/app/(shell)/agent/[slug]/page.mdx:7]: Is this change because of `semibold`? If not, I would stick to markdown for consistency.

lefthook.yml
```diff
@@ -6,9 +6,9 @@ pre-commit:
 
   commands:
     format:
-      glob: "*.{js,jsx,ts,tsx,json,jsonc,css,html,md,mdx,yaml,yml,toml}"
+      glob: "**/*.{js,jsx,ts,tsx,json,jsonc,css,html,yaml,yml,toml}"
```
turisanapo, line 9: Why are we excluding MD and MDX files from being formatted?

packages/ui/src/_mintlify/code/CodeBlock.tsx
```diff
@@ -30,7 +30,7 @@ export const CodeBlock = forwardRef(function CodeBlock(
   const Button = (
     props: Partial<ComponentPropsWithoutRef<typeof CopyToClipboardButton>>,
   ) => (
-    /* FIXME: getNodeText is not picking 100% of text since some nodes are React promises */
+    /* TODO: getNodeText is not picking 100% of text since some nodes are React promises */
```
turisanapo, line 33: Since this won't be solved in this PR, shouldn't it be a FUTURE?

apps/app/src/app/(shell)/agent/[slug]/page.mdx
Comment (heiwen) [reply apps/app/src/app/(shell)/agent/[slug]/page.mdx:7]: It's because of "text-base" - slightly larger font size for the greeting. Couldn't find a markdown equivalent.

lefthook.yml
heiwen: The formatter did a bad job and broke my code samples.

packages/ui/src/_mintlify/code/CodeBlock.tsx
heiwen: The other way around, it's clearly a bug and ideally should be solved before merging. But since the bug was not introduced in this PR, I'm just marking it correctly so it can be handled in a different PR.

lefthook.yml
turisanapo: Is it possible to customise the behaviour of the formatter rather than excluding these files from being automatically formatted? Mostly for the MDX, which are actual code pages.

heiwen: Couldn't find a way.

turisanapo (approved):

PR #50, opened 2025-08-19
implement playground sidebar

Waazify:
## Summary by CodeRabbit

- New Features
  - Right "Playground" sidebar with integrated chat tied to the active branch and branch-aware routing/redirects.

- UI
  - Dual sidebars (left/right) with cookie-based default open states and a Playground trigger showing a keyboard shortcut hint.
  - Enhanced chat input and scrolling behavior plus new chat UI building blocks.

- Data
  - Updated supported model identifiers and display names.

- Chores
  - Introduced a new UI package, updated Tailwind imports, added licenses, and minor formatting/mocks adjustments.


packages/aikit-ui/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "name": "@hebo/aikit-ui",
+  "version": "0.0.1",
+  "private": true,
+  "sideEffects": false,
+  "type": "module",
+  "scripts": {
+    "lint": "bunx eslint .",
+    "clean": "rimraf .turbo node_modules"
+  },
+  "dependencies": {
```
heiwen, line 14: Cleanup dependencies vs devDependencies vs peerDependencies. Use packages/ui as role model.

apps/console/app/routes/_shell/route.tsx
```diff
@@ -89,6 +90,37 @@ export default function ShellLayout({loaderData}: Route.ComponentProps) {
...
+        {/* Floating trigger */}
+        <SidebarTrigger
+          icon={<SquareChevronRight size={24} color="black" />}
+          text="Playground"
+          className="fixed top-4 right-2 z-50 w-fit"
+        />
+
+        <Sidebar side="right" collapsible="offcanvas">
+          <SidebarContent>
+            <Chat models={supportedModels.map((model) => ({
```
heiwen: The interface for the <Chat /> component should be of type model.json.schema. For now, it should only pass a single default model with GROQ as provider and respective API key read from .env.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,304 @@
...
+
+  // Accessibility refs
+  const textareaRef = useRef<HTMLTextAreaElement>(null);
+  const latestMessageRef = useRef<HTMLDivElement>(null);
+
+  const groq = createGroq({
+    apiKey,
+  });
+
+  // Auto-scroll to latest message for screen readers
```
heiwen: Feels weird to introduce a scroll effect specific to "screen readers". Either this is needed for all users, or not needed.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,304 @@
...
+
+  const handleInputChange = useCallback(
+    (e: React.ChangeEvent<HTMLTextAreaElement>) => {
+      setInput(e.target.value);
+    },
+    [],
+  );
+
+  // Enhanced keyboard handling
+  const handleKeyDown = useCallback(
```
heiwen: Would prefer if the keyboard shortcuts are configured one-by-one on the specific elements, instead as a global handler. Would make it much easier to read the code. 

Potentially might make sense to use a separate utility library like https://github.com/JohannesKlauss/react-hotkeys-hook to simplify the code.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,304 @@
...
+  PromptInputModelSelectValue,
+  PromptInputSubmit,
+  PromptInputTextarea,
+  PromptInputToolbar,
+  PromptInputTools,
+} from "@hebo/aikit-ui/_ai-elements/prompt-input";
+
+import { Button } from "../_shadcn/ui/button";
+
+interface Model {
```
heiwen: Please use "type" instead of "interface" for types (entities that only hold data). Interfaces are for entities that have behaviour (i.e. functions).

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,304 @@
...
+  id: string;
+  name: string;
+}
+
+interface ChatProps {
```
heiwen: Please use "type" instead of "interface" for types (entities that only hold data). Interfaces are for entities that have behaviour (i.e. functions).

packages/aikit-ui/src/vite-env.d.ts
heiwen, line 1: This logic should live in apps/console.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,304 @@
...
+  PromptInputModelSelectItem,
+  PromptInputModelSelectTrigger,
+  PromptInputModelSelectValue,
+  PromptInputSubmit,
+  PromptInputTextarea,
+  PromptInputToolbar,
+  PromptInputTools,
+} from "@hebo/aikit-ui/_ai-elements/prompt-input";
+
+import { Button } from "../_shadcn/ui/button";
```
heiwen, line 27: Avoid relative imports. Use "@hebo-aikit-ui" instead like above.

packages/aikit-ui/src/_ai-elements/conversation.tsx
```diff
@@ -0,0 +1,62 @@
+'use client';
+
+import { Button } from "../_shadcn/ui/button";
```
heiwen, line 3: Avoid relative imports. Use "@hebo-aikit-ui" instead like below.

packages/aikit-ui/src/_ai-elements/message.tsx
```diff
@@ -0,0 +1,65 @@
+import {
+  Avatar,
+  AvatarFallback,
+  AvatarImage,
+} from '../_shadcn/ui/avatar';
```
heiwen, line 5: Avoid relative imports. Use "@hebo-aikit-ui" instead like below.

packages/aikit-ui/src/_ai-elements/prompt-input.tsx
```diff
@@ -0,0 +1,247 @@
...
+import {
+  Select,
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from "@hebo/aikit-ui/_shadcn/ui/select";
+import { Textarea } from "@hebo/aikit-ui/_shadcn/ui/textarea";
+import { cn } from "@hebo/aikit-ui/lib/utils";
+import { Button } from "../_shadcn/ui/button";
```
heiwen, line 16: Avoid relative imports. Use "@hebo-aikit-ui" instead like above.

heiwen (commented):
Making good progress.

Some additional comments:

1/ Key thing that needs to be changed is proper separation of concerns. The app/console is in charge of parsing .env files to extract API keys and passing an instance of models.json.schema to the <Chat /> component from packages/aikit-ui. As a library, packages/aikit-ui should never read environment variables itself. 

2/ Avoid excessive useCallback usage for every function call. Only use it when function identity matters (e.g. in memoized children or dependencies).

3/ Don't change the content of any file in _shadcn. If they need small helper utils / hooks, keep them duplicated across packages.

packages/aikit-ui/src/blocks/Chat.tsx
Waazify: Okay I will remove it.

apps/console/vite-env.d.ts
heiwen, line 1: Unnecessary file, delete.

apps/console/app/routes/_shell/route.tsx
```diff
@@ -84,11 +86,51 @@ export default function ShellLayout({loaderData}: Route.ComponentProps) {
...
+          {/* Floating trigger */}
+          <SidebarTrigger
+            icon={<SquareChevronRight size={24} color="black" />}
+            text="Playground"
+            className="fixed top-4 right-2 z-50 w-fit"
+          />
+
+          <Sidebar side="right" collapsible="offcanvas">
+            <SidebarContent>
+              <Chat modelsConfig={{
```
heiwen: 1/ Extract into sidebar-playground component in the route folder.
2/ Don't hardcode values, read from the loaderData.activeAgent instead.
3/ Cleanup unused imports.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,330 @@
...
+    setTimeout(() => textareaRef.current?.focus(), 100);
+  };
+
+  const handleModelChange = (modelAlias: string) => {
+    setCurrentModelAlias(modelAlias);
+    // Return focus to textarea after model change
+    setTimeout(() => textareaRef.current?.focus(), 100);
+  };
+
+  // Auto-focus when Chat becomes visible
```
heiwen: Don't auto-focus. Discouraged by accessibility guidelines.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+type ModelsConfig = {
+  __supportedTypes: string[];
+  models: ModelConfig[];
+};
+
+type ChatProps = { modelsConfig: ModelsConfig };
+
+export function Chat({ modelsConfig }: ChatProps) {
+  // Use the first (and currently only) model as default
```
heiwen: A bit unclear based on your comment whether the code supports only one model or multiple right now?

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,330 @@
...
+          onChange={handleInputChange}
+          onKeyDown={(e) => {
+            if (e.key === "Escape") {
+              if (input) {
+                setInput("");
+              } else {
+                textareaRef.current?.blur();
+              }
+            }
+            if ((e.ctrlKey || e.metaKey) && e.key.toLowerCase() === "r") {
```
heiwen: Why is the reset action bound to the InputTextArea instead of the ResetButton?

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,330 @@
...
+  PromptInputTextarea,
+  PromptInputToolbar,
+  PromptInputTools,
+} from "@hebo/aikit-ui/_ai-elements/prompt-input";
+import { Button } from "@hebo/aikit-ui/_shadcn/ui/button";
+
+// Types based on models.schema.json
+type ModelEndpoint = {
+  baseUrl: string;
+  provider: "aws" | "custom" | "openai";
```
heiwen: Don't define things you are not using.

packages/aikit-ui/src/blocks/Chat.tsx
Waazify: It's actually bound to both. Added to the textarea so the user can reset the chat and keep typing. I think it gives the user a good UX. Should I unbind it from the textarea?

Waazify: I unbound it from the text area and just kept it bound to the reset button.

apps/console/app/mocks/agents.ts
heiwen, line 1: Revert changes and only apply bare minimum that you need for your added functionality.

In general, the mock API implementation details do not have to match the real API (otherwise you can just use the real API). It's a simplified version that helps develop the frontend while the API is not ready.

For example, if you never expose an ID, there is no need to implement one internally. The slug is already marked as unique in the schema. Changing that just adds a lot of new boilerplate.

apps/console/app/mocks/branches.ts
heiwen, line 1: Revert changes and only apply bare minimum that you need for your added functionality.

In general, the mock API implementation details do not have to match the real API (otherwise you can just use the real API). It's a simplified version that helps develop the frontend while the API is not ready.

For example, there is no need for a MockBranch type and cleanBranchResponse functionality, as this is just casting between a never used internal representation and an external representation. Don't add anything that is not exposed to the internal representation in the first place.

apps/console/app/mocks/db.ts
heiwen, line 1: Revert changes and only apply bare minimum that you need for your added functionality.

In general, the mock API implementation details do not have to match the real API (otherwise you can just use the real API). It's a simplified version that helps develop the frontend while the API is not ready.

For example, why would you need both an agentId and an agent field.

apps/console/app/routes/_shell/route.tsx
```diff
@@ -22,42 +22,53 @@ import type { Route } from "./+types/route";
...
+  if (!activeAgent) throw new Response("Agent Not Found", { status: 404 });
+  
+  const { data: branches = [] } = await api.agents[params.slug].branches.get();
+  
+  return { 
+    agents, 
+    activeAgent, 
+    branches, 
+    activeBranch: branches.find((b: any) => b.slug === "main") || branches[0]
+  };
```
heiwen: Revert changes, follow structure of the original code, and just add what you need.

apps/console/app/routes/_shell/route.tsx
```diff
@@ -22,42 +22,53 @@ import type { Route } from "./+types/route";
...
+  const activeAgent = agents?.find((a: any) => a.slug === params.slug);
+  if (!activeAgent) throw new Response("Agent Not Found", { status: 404 });
+  
+  const { data: branches = [] } = await api.agents[params.slug].branches.get();
+  
+  return { 
+    agents, 
+    activeAgent, 
+    branches, 
+    activeBranch: branches.find((b: any) => b.slug === "main") || branches[0]
```
heiwen: Don't hardcode "main" here. Instead,
1/ It should be represented by an additional path parameter for the branch slug
2/ Default redirect (i.e. select first branch) to be set in _shell.index/route.tsx.

apps/console/app/routes/_shell/route.tsx
```diff
@@ -84,11 +95,48 @@ export default function ShellLayout({loaderData}: Route.ComponentProps) {
         <main className="relative flex w-full flex-1 flex-col gap-4 p-4">
           <SidebarTrigger className="fixed -m-1.5" />
 
-          <div className="mx-auto flex w-full max-w-4xl min-w-0 flex-col gap-2 py-8">
+          <div className="flex-1 mx-auto flex w-full max-w-4xl min-w-0 flex-col gap-2 p-4 py-8">
```
heiwen: What is this change for?

apps/console/app/routes/_shell/route.tsx
```diff
@@ -84,11 +95,48 @@ export default function ShellLayout({loaderData}: Route.ComponentProps) {
...
+          }
+        >
+          {/* Floating trigger */}
+          <SidebarTrigger
+            icon={
+              <div className="flex items-center gap-2">
+                <SquareChevronRight size={24} color="black" />
+                <span>Playground</span>
+                <span className="text-muted-foreground">
+                  {/Mac|iPhone|iPad|iPod/.test(navigator.userAgent) ? '⌘P' : 'Ctrl P'}
```
heiwen: These values should not be hardcoded in a component. Create helper util instead. There are probably existing libraries that do this.

apps/console/app/routes/_shell/sidebar-playground.tsx
```diff
@@ -0,0 +1,36 @@
...
+export function PlaygroundSidebar({ activeBranch }: PlaygroundSidebarProps) {
+  if (!activeBranch?.models?.length) {
+    return (
+      <div className="flex items-center justify-center h-full text-muted-foreground">
+        <p>Agent doesn't have a model yet</p>
+      </div>
+    );
+  }
+
+  const modelsConfig = {
```
heiwen, line 19: Instead of creating a new models object, reuse the one that's stored in the branches table and augment baseUrl / apiKey into each model.

 __supportedTypes does not have to be provided, it's just for schema verification.

packages/aikit-ui/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "@ai-sdk/openai": "^2.0.17",
+    "ai": "^5.0.13",
```
heiwen: Should ai-sdk and ai be dependencies or peerDepedencies?

packages/aikit-ui/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "react-dom": "^19.1.1"
+  },
+  "devDependencies": {
+    "@radix-ui/react-avatar": "^1.1.10",
+    "@radix-ui/react-select": "^2.2.5",
+    "@radix-ui/react-slot": "^1.2.3",
+    "@types/node": "^24.3.0",
+    "@types/react": "^19.1.1",
+    "@types/react-dom": "^19.1.1",
+    "ai-elements": "^1.0.3",
```
heiwen: What does this do?

packages/aikit-ui/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "@radix-ui/react-slot": "^1.2.3",
+    "lucide-react": "^0.539.0",
+    "react": "^19.1.1",
+    "react-dom": "^19.1.1"
+  },
+  "devDependencies": {
+    "@radix-ui/react-avatar": "^1.1.10",
+    "@radix-ui/react-select": "^2.2.5",
+    "@radix-ui/react-slot": "^1.2.3",
+    "@types/node": "^24.3.0",
```
heiwen: Do you need this?

packages/aikit-ui/src/_shadcn/ui/select.tsx
```diff
@@ -1,10 +1,8 @@
-"use client"
-
 import * as React from "react"
 import * as SelectPrimitive from "@radix-ui/react-select"
 import { CheckIcon, ChevronDownIcon, ChevronUpIcon } from "lucide-react"
 
-import { cn } from "@hebo/ui/lib/utils"
+import { cn } from "@hebo/aikit-ui/lib/utils"
```
heiwen: We discussed not to change _shadcn files.

heiwen: We discussed not to change _shadcn files.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+  PromptInputModelSelectValue,
+  PromptInputSubmit,
+  PromptInputTextarea,
+  PromptInputToolbar,
+  PromptInputTools,
+} from "@hebo/aikit-ui/_ai-elements/prompt-input";
+import { Button } from "@hebo/aikit-ui/_shadcn/ui/button";
+
+// Types based on models.schema.json
+type ModelEndpoint = {
```
heiwen: This is part of the ModelsConfig, it doesn't exist as a standalone type.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+  baseUrl: string;
+  apiKey: string;
+};
+
+type ModelConfig = {
```
heiwen: This is part of the ModelsConfig, it doesn't exist as a standalone type.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+  type: string;
+  endpoint?: ModelEndpoint;
+};
+
+type ModelsConfig = {
+  __supportedTypes: string[];
+  models: ModelConfig[];
+};
+
+type ChatProps = { modelsConfig: ModelsConfig };
```
heiwen: Unnecessary for a single prop, just inline.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+export function Chat({ modelsConfig }: ChatProps) {
+  // Use lazy initializer to safely pick first model ID or empty string
+  const [currentModelAlias, setCurrentModelAlias] = useState(() =>
```
heiwen: Why lazy?

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+        setTimeout(() => textareaRef.current?.focus(), 100);
+      }
+    },
+    [input, isLoading, messages, currentModel, openai],
+  );
+
+  const handleReset = () => {
+    setMessages([]);
+    // Focus textarea after reset
+    setTimeout(() => textareaRef.current?.focus(), 100);
```
heiwen: Don't do that. Manually setting focus is against accessibility guidelines. Instead, focus on detecting keystrokes (like ChatGPT or Claude do in their web apps).

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+  };
+
+  const handleModelChange = (modelAlias: string) => {
+    setCurrentModelAlias(modelAlias);
+    // Return focus to textarea after model change
+    setTimeout(() => textareaRef.current?.focus(), 100);
```
heiwen: Don't do that. Manually setting focus is against accessibility guidelines. Instead, focus on detecting keystrokes (like ChatGPT or Claude do in their web apps).

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+  const [currentModelAlias, setCurrentModelAlias] = useState(() =>
+    modelsConfig.models.length > 0 ? modelsConfig.models[0].alias : "",
+  );
+  const [messages, setMessages] = useState<UIMessage[]>([]);
+  const [input, setInput] = useState("");
+  const [isLoading, setIsLoading] = useState(false);
+
+  // Accessibility refs
+  const textareaRef = useRef<HTMLTextAreaElement>(null);
+  const containerRef = useRef<HTMLDivElement>(null);
```
heiwen: Is this ever used?

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+  // Use lazy initializer to safely pick first model ID or empty string
+  const [currentModelAlias, setCurrentModelAlias] = useState(() =>
+    modelsConfig.models.length > 0 ? modelsConfig.models[0].alias : "",
+  );
+  const [messages, setMessages] = useState<UIMessage[]>([]);
+  const [input, setInput] = useState("");
+  const [isLoading, setIsLoading] = useState(false);
+
+  // Accessibility refs
+  const textareaRef = useRef<HTMLTextAreaElement>(null);
```
heiwen: Give it a more specific name.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+  const renderMessagePart = (part: UIMessage["parts"][0]) => {
+    if (part.type === "text") return part.text;
+    if (part.type === "dynamic-tool" && "input" in part) {
+      return JSON.stringify(part.input);
+    }
+    return "";
+  };
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
```
heiwen: One-liner can be inlined.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+      {/* Header Controls */}
+      <div className="absolute top-4 left-4 z-10 flex items-center">
+        <Button
+          variant="ghost"
+          size="icon"
+          onClick={handleReset}
+          aria-label="Clear conversation"
+          title="Clear conversation"
+        >
+          <IterationCcw size={20} color="#525252" />
```
heiwen: No hardcoded color values.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+          </PromptInputTools>
+
+          {/* Attachment button */}
+          <PromptInputButton
+            className="absolute right-10 bottom-1"
+            disabled={isLoading || !currentModel}
+            aria-label="Attach file"
+            title="Attach file"
+          >
+            <PaperclipIcon size={16} color="#525252" />
```
heiwen: No hardcoded color values.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+          </PromptInputButton>
+
+          {/* Submit button - disable when no model is selected */}
+          <PromptInputSubmit
+            disabled={!input.trim() || isLoading || !currentModel}
+            className="absolute right-1 bottom-1 text-[#525252]"
```
heiwen: No hardcoded color values.

packages/aikit-ui/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "baseUrl": "src",
+    "paths": {
+      "@hebo/aikit-ui/*": ["./*"]
+    }
+  },
+  "include": ["src", "src/vite-env.d.ts"],
```
heiwen: There should be no vite-env in the aikit-ui.

packages/ui/components.json
```diff
@@ -5,17 +5,17 @@
...
-    "css": "../../apps/app/src/styles/tailwind.css",
+    "css": "../../apps/console/app/styles/tailwind.css",
     "baseColor": "neutral",
     "cssVariables": true,
     "prefix": ""
   },
   "aliases": {
     "components": "@hebo/ui/_shadcn",
-    "utils": "@hebo/ui/lib/utils",
+    "utils": "@hebo/aikit-ui/lib/utils",
```
heiwen: Revert

packages/ui/components.json
```diff
@@ -5,17 +5,17 @@
...
     "ui": "@hebo/ui/_shadcn/ui",
-    "lib": "@hebo/ui/lib",
-    "hooks": "@hebo/ui/hooks"
+    "lib": "@hebo/aikit-ui/lib",
```
heiwen: Revert

packages/ui/components.json
```diff
@@ -5,17 +5,17 @@
...
+    "hooks": "@hebo/aikit-ui/hooks"
```
heiwen: Revert

packages/ui/src/_shadcn/ui/sidebar.tsx
heiwen, line 1: Do not modify _shadcn components

packages/ui/src/components/Button.tsx
```diff
@@ -3,8 +3,8 @@ import { Loader2Icon } from "lucide-react";
 import {
   Button as ShadCNButton,
   buttonVariants,
-} from "@hebo/ui/_shadcn/ui/button";
-import { cn } from "@hebo/ui/lib/utils";
+} from "@hebo/aikit-ui/_shadcn/ui/button";
+import { cn } from "@hebo/aikit-ui/lib/utils";
```
heiwen: Use cn from ui package

packages/ui/src/components/Input.tsx
```diff
@@ -1,6 +1,6 @@
+import { cn } from "@hebo/aikit-ui/lib/utils";
```
heiwen: Use cn from ui package

apps/console/app/routes/_shell/route.tsx
```diff
@@ -84,11 +95,48 @@ export default function ShellLayout({loaderData}: Route.ComponentProps) {
...
 
-          <div className="mx-auto flex w-full max-w-4xl min-w-0 flex-col gap-2 py-8">
+          <div className="flex-1 mx-auto flex w-full max-w-4xl min-w-0 flex-col gap-2 p-4 py-8">
             <Outlet />
           </div>
         </main>
       </SidebarProvider>
+      
+      {/* RIGHT SIDEBAR - Positioned absolutely */}
+      <div className="absolute top-0 right-0 h-full z-40">
```
heiwen: Use flex positioning instead of absolute. See shadcn sidebar components as example.

packages/aikit-ui/src/_ai-elements/conversation.tsx
heiwen, line 1: Don't forget to add LICENSE file for _ai_elements

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+        const errorMessage: UIMessage = {
+          id: crypto.randomUUID(),
+          role: "assistant",
+          parts: [{ type: "text", text: "Sorry, I encountered an error." }],
+        };
+        setMessages((prev) => [...prev, errorMessage]);
+      } finally {
+        setIsLoading(false);
+        // Return focus to textarea after response
+        setTimeout(() => textareaRef.current?.focus(), 100);
```
heiwen: Don't do that. Manually setting focus is against accessibility guidelines. Instead, focus on detecting keystrokes (like ChatGPT or Claude do in their web apps).

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+        onSubmit={handleSubmit}
+        className="relative mt-4"
+        role="form"
+        aria-label="Chat input form"
+      >
+        <PromptInputTextarea
+          ref={textareaRef}
+          id="chat-input"
+          onChange={handleInputChange}
+          onKeyDown={(e) => {
```
heiwen: Unnecessary, nobody will ever use that. Remove.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+              <PromptInputModelSelect
+                onValueChange={handleModelChange}
+                value={currentModelAlias}
+                disabled={isLoading || modelsConfig.models.length === 0}
+                aria-label="Select AI model"
+              >
+                <PromptInputModelSelectTrigger
+                  aria-label={`Current model: ${currentModel?.alias || "None"}`}
+                >
+                  <Bot color="#525252" />
```
heiwen: No hardcoded color values.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+        setIsLoading(false);
+        // Return focus to textarea after response
+        setTimeout(() => textareaRef.current?.focus(), 100);
+      }
+    },
+    [input, isLoading, messages, currentModel, openai],
+  );
+
+  const handleReset = () => {
+    setMessages([]);
```
heiwen: One-liner can be inlined.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+    // Focus textarea after reset
+    setTimeout(() => textareaRef.current?.focus(), 100);
+  };
+
+  const handleModelChange = (modelAlias: string) => {
+    setCurrentModelAlias(modelAlias);
```
heiwen: One-liner can be inlined.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+    return "";
+  };
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
+    setInput(e.target.value);
+  };
+
+  // No global keyboard handler; attach element-specific handlers below
+
+  const handleSubmit = useCallback(
```
heiwen: Unnecessary useCallback.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+            {/* Model selector - show hint when no models available */}
+            {modelsConfig.models.length === 0 ? (
+              <div className="text-muted-foreground px-3 py-2 text-sm">
+                No models available
+              </div>
+            ) : (
+              <PromptInputModelSelect
+                onValueChange={handleModelChange}
+                value={currentModelAlias}
+                disabled={isLoading || modelsConfig.models.length === 0}
```
heiwen: This code never executes when modelsConfig.models.length === 0. Check can be removed.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+                aria-label="Select AI model"
+              >
+                <PromptInputModelSelectTrigger
+                  aria-label={`Current model: ${currentModel?.alias || "None"}`}
```
heiwen: Alias should never be undefined, should it?

apps/console/app/routes/_shell/sidebar-playground.tsx
```diff
@@ -0,0 +1,36 @@
...
+    type: string;
+  }>;
+};
+
+interface PlaygroundSidebarProps {
+  activeBranch?: Branch;
+}
+
+export function PlaygroundSidebar({ activeBranch }: PlaygroundSidebarProps) {
+  if (!activeBranch?.models?.length) {
```
heiwen: This will never happen in reality, remove.

apps/console/app/routes/_shell/sidebar-playground.tsx
```diff
@@ -0,0 +1,36 @@
+import { Chat } from "@hebo/aikit-ui/blocks/Chat";
+
+type Branch = {
+  models?: Array<{
+    alias: string;
+    type: string;
+  }>;
+};
+
+interface PlaygroundSidebarProps {
```
heiwen: Unnecessary definition for a single item, just inline.

apps/console/app/routes/_shell/route.tsx
```diff
@@ -84,11 +95,48 @@ export default function ShellLayout({loaderData}: Route.ComponentProps) {
...
+            icon={
+              <div className="flex items-center gap-2">
+                <SquareChevronRight size={24} color="black" />
+                <span>Playground</span>
+                <span className="text-muted-foreground">
+                  {/Mac|iPhone|iPad|iPod/.test(navigator.userAgent) ? '⌘P' : 'Ctrl P'}
+                </span>
+              </div>
+            }
+            className="fixed top-4 right-2 z-50 w-fit"
```
heiwen: z-index and w-fit should not be necessary.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,300 @@
...
+  const handleModelChange = (modelAlias: string) => {
+    setCurrentModelAlias(modelAlias);
+    // Return focus to textarea after model change
+    setTimeout(() => textareaRef.current?.focus(), 100);
+  };
+
+  return (
+    <div className="flex h-full flex-col" ref={containerRef}>
+      {/* Header Controls */}
+      <div className="absolute top-4 left-4 z-10 flex items-center">
```
heiwen, line 141: Avoid usage of z-values unless really necessary.

packages/aikit-ui/src/blocks/Chat.tsx
Waazify: Because React evaluates useState’s argument on every render, a heavy/derived initial value would be recomputed each time. Passing a function (lazy initializer) defers that computation until the first render only, avoiding unnecessary work on re-renders and making it safe if the expression could be expensive or reference data that might not be stable during render.

Example:
Without lazy: useState(modelsConfig.models[0].alias) runs that lookup every render.
With lazy: useState(() => modelsConfig.models[0].alias) runs it once (initial mount).

packages/aikit-ui/src/blocks/Chat.tsx
Waazify: In this case it really is necessary. Because the conversation element is rendering the ChatBubbles on a higher z-index and for the the Reset button in the header controls to be usable it needs to be on a higher z-index to be usable.

packages/aikit-ui/package.json
Waazify: ai-elements is the package we use to fetch ai-elements and put them in our source code. Since we only used it in development i put it in devDeps. If we are not planning to use it anymore (to fetch ai-elements) we can remove it from devDependencies all together.

packages/aikit-ui/src/blocks/Chat.tsx
Waazify: Yes. It is assigned as the ref to the parent container.

packages/aikit-ui/src/blocks/Chat.tsx
heiwen: Yes, but (i) it's not heavy (ii) re-renders will be rare.

packages/aikit-ui/src/blocks/Chat.tsx
heiwen: And what's the purpose of that assignment?

apps/console/app/routes/_shell/route.tsx
Waazify: z-index was removed, however, the w-fit is necessary because the SidebarTrigger has a weird behaviour where it forces the side of the icon to be 9 (width x height) and the only way to override it is to use w-fit.

apps/console/app/lib/utils.ts
```diff
@@ -6,3 +6,17 @@ export function getCookie(name: string): string | undefined {
...
   }
 }
+
+/**
+ * Returns the appropriate keyboard shortcut text based on the user's platform
+ * @param shortcut - The shortcut key (e.g., 'P' for the P key)
+ * @returns The formatted shortcut string (e.g., '⌘P' on Mac, 'Ctrl P' on other platforms)
+ */
+export const getKeyboardShortcut = (shortcut: string): string => {
+  if (typeof navigator === "undefined") {
```
heiwen: Why would this ever happen? Doesn't your normal return statement already cover for this?

apps/console/app/lib/utils.ts
```diff
@@ -6,3 +6,17 @@ export function getCookie(name: string): string | undefined {
     return value?.split("=")[1];
   }
 }
+
+/**
+ * Returns the appropriate keyboard shortcut text based on the user's platform
+ * @param shortcut - The shortcut key (e.g., 'P' for the P key)
+ * @returns The formatted shortcut string (e.g., '⌘P' on Mac, 'Ctrl P' on other platforms)
+ */
+export const getKeyboardShortcut = (shortcut: string): string => {
```
heiwen: It will be frequently used in code, suggest to shorten it to something like "kbs(...)".

apps/console/app/mocks/agents.ts
```diff
@@ -5,10 +5,22 @@ import { db } from "~/mocks/db";
 
 export const agentHandlers = [
   http.post("/api/v1/agents", async ({ request }) => {
-    const body = (await request.json()) as ReturnType<typeof db.agent.create>;
+    const body = (await request.json()) as {
```
heiwen: Can't you keep the original return value derived from create?

apps/console/app/routes/_shell/route.tsx
```diff
@@ -14,50 +14,61 @@ import {
...
 
 export const unstable_clientMiddleware = [authMiddleware];
 
-
 export async function clientLoader({ params }: Route.ClientLoaderArgs) {
   const { data: agents } = await api.agents.get();
-
-  const activeAgent = params.slug ? agents!.find((a) => a.slug === params.slug) : undefined;
+  
+  const activeAgent = params.slug ? agents!.find((a: any) => a.slug === params.slug) : undefined;
```
heiwen: Why are you introducing a new "any" here?  React Router is typed by default via ClientLoaderArgs.

apps/console/app/routes/_shell/route.tsx
```diff
@@ -14,50 +14,61 @@ import {
...
+  
+  // Add branch slug identification
+  const activeBranch = (params as any).branchSlug && activeAgent
```
heiwen: We don't use "any". React Router is typed by default via ClientLoaderArgs.

apps/console/app/routes/_shell/route.tsx
```diff
@@ -14,50 +14,61 @@ import {
...
+  
+  // Add branch slug identification
+  const activeBranch = (params as any).branchSlug && activeAgent 
+    ? activeAgent.branches?.find((b: any) => b.slug === (params as any).branchSlug)
+    : undefined;
 
   if (params.slug && !activeAgent)
     throw new Response("Agent Not Found", { status: 404 });
+    
+  if ((params as any).branchSlug && !activeBranch)
```
heiwen: Remove any. React Router is typed by default via ClientLoaderArgs

apps/console/app/routes/_shell/route.tsx
```diff
@@ -14,50 +14,61 @@ import {
...
-  const activeAgent = params.slug ? agents!.find((a) => a.slug === params.slug) : undefined;
+  
+  const activeAgent = params.slug ? agents!.find((a: any) => a.slug === params.slug) : undefined;
+  
+  // Add branch slug identification
+  const activeBranch = (params as any).branchSlug && activeAgent 
+    ? activeAgent.branches?.find((b: any) => b.slug === (params as any).branchSlug)
+    : undefined;
 
   if (params.slug && !activeAgent)
```
heiwen: Since you are introducing "branchSlug", you should rename "slug" to "agentSlug" everywhere to avoid ambiguity.

apps/console/app/routes/_shell/route.tsx
```diff
@@ -14,50 +14,61 @@ import {
...
   const { user } = useSnapshot(authStore);
 
   // FUTURE replace with session storage
-  const defaultOpen = getCookie("sidebar_state") === "true";
+  const leftSidebarDefaultOpen = getCookie("left_sidebar_state") === "true";
+  const rightSidebarDefaultOpen = getCookie("right_sidebar_state") === "true";
 
   return (
-    <div className="flex min-h-screen flex-col gap-4">
+    <div className="relative flex min-h-screen">
```
heiwen: Why this change?

apps/console/app/routes/_shell/sidebar-playground.tsx
```diff
@@ -0,0 +1,25 @@
...
+type Branch = {
+  models?: Array<{
+    alias: string;
+    type: string;
+  }>;
+};
+
+export function PlaygroundSidebar({ activeBranch }: { activeBranch?: Branch }) {
+  const gatewayBaseUrl = String(import.meta.env.VITE_GATEWAY_URL ?? "");
+  const gatewayApiKey = String(import.meta.env.VITE_GATEWAY_API_KEY ?? "");
```
heiwen: API_KEY is not required, JWT authentication will be used between console and gateway (same as between console and api).

apps/console/app/routes/_shell/sidebar-playground.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+type Branch = {
+  models?: Array<{
+    alias: string;
+    type: string;
+  }>;
+};
+
+export function PlaygroundSidebar({ activeBranch }: { activeBranch?: Branch }) {
+  const gatewayBaseUrl = String(import.meta.env.VITE_GATEWAY_URL ?? "");
```
heiwen: You are hiding an error condition (aka VITE_GATEWAY_URL is not set). Log an error.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,250 @@
...
+export function Chat({ modelsConfig }: { modelsConfig: ModelsConfig }) {
+  const [currentModelAlias, setCurrentModelAlias] = useState(
+    modelsConfig.models.length > 0 ? modelsConfig.models[0].alias : "",
+  );
+  const [messages, setMessages] = useState<UIMessage[]>([]);
+  const [input, setInput] = useState("");
+  const [isLoading, setIsLoading] = useState(false);
+
+  // Accessibility refs
+  const promptInputTextareaRef = useRef<HTMLTextAreaElement>(null);
```
heiwen: What is this used / needed for?

packages/aikit-ui/src/blocks/Chat.tsx
heiwen, line 1: I'm looking for the part where typing focuses the PromptInputTextArea (like in ChatGPT / Claude web interfaces)?

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,250 @@
...
+      <PromptInput
+        onSubmit={handleSubmit}
+        className="relative mt-4"
+        role="form"
+        aria-label="Chat input form"
+      >
+        <PromptInputTextarea
+          ref={promptInputTextareaRef}
+          id="chat-input"
+          onChange={(e) => setInput(e.target.value)}
```
heiwen, line 200: Consider debouncing this

heiwen (commented):
Almost there, set of hopefully last comments.

apps/console/app/routes/_shell/route.tsx
Waazify: There was a UI bug where when we added the Playground component we would get a weird scrollable gap after the main content. This was to address that bug.

apps/console/app/routes/_shell._index/route.tsx
```diff
@@ -6,10 +6,22 @@ async function defaultAgentMiddleware() {
...
+    throw redirect("/agent/create");
   }
+
+  try {
+    const { data: branches } = await api.agents({ agentSlug: agents[0].slug }).branches.get();
+    
+    if (branches?.length) {
+      throw redirect(`/agent/${agents[0].slug}/branch/${branches[0].slug}`);
+    }
+  } catch {
```
heiwen: Don't hide errors. API errors just bubble to the top and will be handled by a global error handler in the future.

apps/console/app/routes/_shell._index/route.tsx
```diff
@@ -6,10 +6,22 @@ async function defaultAgentMiddleware() {
   const { data: agents } = await api.agents.get();
 
   // FUTURE fade-in the next page
-  // FUTURE remember last agent in sessions storage
-  if (agents!.length > 0) {
-    throw redirect(`/agent/${agents![0]?.slug}`);
+  // FUTURE remember last agent and branch in session storage
+  
+  if (!agents?.length) {
```
heiwen: Agents should never be undefined, that would be an API error. Please revert to previous version.

apps/console/app/routes/_shell._index/route.tsx
```diff
@@ -6,10 +6,22 @@ async function defaultAgentMiddleware() {
...
+    throw redirect("/agent/create");
   }
+
+  try {
+    const { data: branches } = await api.agents({ agentSlug: agents[0].slug }).branches.get();
+    
+    if (branches?.length) {
```
heiwen: Branches should never be undefined, that would be an API error. Please match previous agent version.

apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug/route.tsx
heiwen, line 1: Please move _shell.agent.$agentSlug (which is no longer used after this change), instead of recreating the route layout.

apps/console/app/routes/_shell/route.tsx
```diff
@@ -14,50 +14,62 @@ import {
...
+    throw new Response("Branch Not Found", { status: 404 });
 
-  return { agents, activeAgent };
+  return { agents, activeAgent, activeBranch };
 }
 
 export function shouldRevalidate({ currentParams, nextParams }: ShouldRevalidateFunctionArgs) {
-  // Only reload data if the slug exists and changed
-  return nextParams.slug !== undefined && currentParams.slug !== nextParams.slug;
+  // Update to use agentSlug instead of slug
```
heiwen: Revert comment to previous one

apps/console/app/routes/_shell/route.tsx
heiwen: I'm not convinced this is the right fix, can you elaborate more?

apps/console/app/routes/_shell/sidebar-agent.tsx
```diff
@@ -97,7 +97,7 @@ export function AgentSelect({
             <DropdownMenuSeparator />
             {agents.map((agent) => (
               <DropdownMenuItem key={agent.slug} className="gap-2 p-2" asChild>
-                <Link to={`/agent/${agent.slug}`} viewTransition>
+                <Link to={`/agent/${agent.slug}/branch/main`} viewTransition>
```
heiwen: Don't hardcode branch name

apps/console/app/routes/_shell/sidebar-playground.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+type Branch = {
+  models?: Array<{
+    alias: string;
+    type: string;
+  }>;
+};
+
+export function PlaygroundSidebar({ activeBranch }: { activeBranch?: Branch }) {
+  const gatewayBaseUrl = import.meta.env.VITE_GATEWAY_URL?.toString().trim();
```
heiwen: no toString() needed

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,288 @@
...
+  // Create OpenAI client based on current model (only if model exists)
+  const openai = currentModel
+    ? createOpenAI({
+        apiKey: "",
+        baseURL: currentModel.endpoint?.baseUrl || "",
+        fetch: currentModel.endpoint?.fetch,
+      })
+    : undefined;
+
+  useEffect(() => {
```
heiwen, line 72: Why are you re-introducing useEffect() here?

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,288 @@
...
+  const renderMessagePart = (part: UIMessage["parts"][0]) => {
+    if (part.type === "text") return part.text;
+    if (part.type === "dynamic-tool" && "input" in part) {
+      return JSON.stringify(part.input);
+    }
+    return "";
+  };
+
+  useEffect(() => {
+    const handleGlobalKeyDown = (e: KeyboardEvent) => {
```
heiwen: I thought we discussed not to do this?

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,288 @@
...
+      // Don't focus if user is already typing in an input
+      if (
+        e.target instanceof HTMLInputElement ||
+        e.target instanceof HTMLTextAreaElement ||
+        (e.target instanceof HTMLElement && e.target.isContentEditable)
+      ) {
+        return;
+      }
+
+      // Only trigger on Ctrl+L (Windows/Linux) or Cmd+L (Mac)
```
heiwen, line 101: Move shortcuts to the object instead of global handler (as discussed previously)

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,272 @@
+"use client";
+
+import { createOpenAI } from "@ai-sdk/openai";
+import { generateText, type UIMessage } from "ai";
```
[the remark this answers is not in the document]
Waazify: We want to use the Client Side SDK for now. No need to keep bringing it up.

packages/aikit-ui/src/blocks/Chat.tsx
```diff
@@ -0,0 +1,272 @@
...
+            fetch(input, {
+              ...init,
+              credentials: "include",
+            })),
+      })
+    : undefined;
+
+  useEffect(() => {
+    const handleGlobalKeyDown = (e: KeyboardEvent) => {
+      // Skip if already in an input
```
heiwen: Why? I could be in a totally different input and want to jump to the playground chat.

Waazify: Understood.  I updated it to remove the part where it skips the shortcut if it's already in another input.

packages/aikit-ui/src/blocks/Chat.tsx
Waazify: I thought about it. 

I think debouncing makes sense when:

- You're making API calls on every keystroke
- Doing expensive computations/filtering on large datasets
- The operation has noticeable performance impact

But this is a simple and feels a little overkill to add debouncing.

packages/aikit-ui/components.json
```diff
@@ -0,0 +1,21 @@
...
+  "rsc": true,
+  "tsx": true,
+  "tailwind": {
+    "config": "",
+    "css": "../../apps/console/app/styles/tailwind.css",
+    "baseColor": "neutral",
+    "cssVariables": true,
+    "prefix": ""
+  },
+  "aliases": {
```
heiwen, line 13: Change to use relative paths, as per new standard. Have a look at ui package.

packages/ui/src/_shadcn/ui/textarea.tsx
heiwen, line 1: I don't think this is needed anymore since it is only used in aikit right now.

packages/aikit-ui/src/blocks/Chat.tsx
heiwen: Makes sense

heiwen (commented):
LGTM. Only minor artefacts remaining from rebase.

packages/aikit-ui/components.json
Waazify: image
Do you think this comment is valid? If it is then the relative imports for the ui package needs to be updated as well.

heiwen: You can figure that out yourself. Does adding shadcn components with this configuration work correctly or do you have to manually modify the imports in the components after adding?

Waazify: With these current imports it doesn't work. I will address according to the comment suggestion, both aikit-ui and ui.

turisanapo (approved):

PR #51, opened 2025-08-19
Request-Scoped DB & Audit Refactor

turisanapo:
## Summary by CodeRabbit

* **New Features**
  * Per-request DB transactions, request-scoped auditing, and user-id middleware; stricter agent model validation.
  * DB-level lowercase slug checks and soft-delete support.

* **Bug Fixes**
  * Prevents slug collisions with soft-deleted records.

* **Refactor**
  * Replaced legacy audit middleware with user-id flow; services updated to accept userId and use request-scoped DB access.

* **Chores**
  * Added request logging middleware and improved dev log formatting; package updates for logging.


apps/api/src/middlewares/user-id.ts
heiwen, line 1: I assume that this middleware will be replaced by the auth middleware later?

apps/api/src/modules/agents/index.ts
```diff
@@ -9,14 +10,14 @@ export const agentsModule = new Elysia({
   name: "agents-module",
   prefix: "/agents",
 })
-  .use(auditFields)
+  .use(userId)
   .post(
     "/",
-    async ({ body, set, auditFields }) => {
-      const agent = await AgentService.createAgent(body, auditFields);
+    withRequestTransaction(async ({ body, set, userId }) => {
```
heiwen: 1/ This breaks conceptual dependency boundaries. Essentially you are saying the route needs to know that the service needs a transaction (whereas all the other's don't). The route doesn't know that. The wrapper should either be within the service instead, or all routes needs to be wrapped with a transaction. Prefer the first, since it gives more control on the lower level.

2/ How does this actually work? Does this automatically create a new transaction that is only finished when the requests ends? Or how does it know about transaction boundaries?

3/ Should the agent-id middleware be part of the same transaction or does it need to be separate? If it's separate, should it still reuse the same db object?

packages/db/drizzle.ts
```diff
@@ -69,3 +69,4 @@ export const db = initDb();
 export const typedDb: UniversalDb = db;
 
 export { isLocal } from "./utils";
+export * from "./audit-utils";
```
heiwen: We agreed to avoid barrel files and use deep imports instead.

Related 1/: Should isLocal be exported at all? Who uses it?

Related 2/: Suggest to update packages.json to use ESM syntax instead of old CommonJs syntax. Meaning use "exports" instead of "main / types". Have a look at packages/ui for reference (although it doesn't export types).

packages/db/audit-utils.ts
heiwen, line 1: move into utils/with-audit.ts

heiwen (commented):
Additional general comments:

1/ Some of the CodeRabbit nitpicks seem valid, especially around the new request-db util.

2/ There are a lot of types defined in {agents/branches}/model.ts, which one of these are we actually using / can be removed?

3/ Right now the services don't correctly interfere the types from the models (during development). If helpful, this can be solved by making the services ElysiaJS instances:

See: "Best Practices - Do: Request dependent service as Elysia instance" (https://elysiajs.com/essential/best-practice.html#service)

Nitpick 1: give the PR a more meanigful name

Nitpick 2: most of the TODOs in the code are not TODOs (blocking from merge) but FUTURE tags.

packages/db/audit-utils.ts
```diff
@@ -0,0 +1,75 @@
...
+    insert(db: UniversalDbClient, values: object) {
+      return (db as any).insert(table as any).values({
+        ...values,
+        createdBy: ctx.userId,
+        updatedBy: ctx.userId,
+        // keep server-side timestamps consistent regardless of client clock
+        createdAt: sql`now()`,
+        updatedAt: sql`now()`,
+      });
+    },
```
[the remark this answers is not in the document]
turisanapo: unnecessary complexity at the moment, considering as well that these columns are added by an audit file helper when tables are created

turisanapo:
> 3/ Right now the services don't correctly interfere the types from the models (during development). If helpful, this can be solved by making the services ElysiaJS instances:
> 
> See: "Best Practices - Do: Request dependent service as Elysia instance" (https://elysiajs.com/essential/best-practice.html#service)

The `any` isn’t coming from Elysia service composition; it’s caused by `withAudit` using `as any`, which erases Drizzle’s type inference, so `.returning()` becomes `any[]` and `[agent]` is inferred as `any`.

I can try to refactor `withAudit` to be fully typed, but it looks like overcomplicating the utility right now. What do you think?

apps/api/src/middlewares/user-id.ts
turisanapo: Correct

apps/api/src/modules/agents/index.ts
turisanapo: 1/ will commit a change soon

2/ The change is based on [Asynchronous context tracking](https://nodejs.org/download/release/v18.17.1/docs/api/async_context.html#asynchronous-context-tracking), whose `AsyncLocalStorage` make the db client available anywhere downstream the wrapper.

`withRequestTransaction` starts a database transaction for the wrapped handler and exposes that transaction via `AsyncLocalStorage` so any code that calls `getDb()` inside that handler automatically uses the same transactional client.

The boundary is the lifetime of the wrapped handler’s Promise. The underlying `baseDb.transaction` commits when the handler resolves and rolls back if it throws/rejects.

It uses `AsyncLocalStorage` to store the active DB client for the current async context. `getDb()` returns that client if present, otherwise falls back to the base client.

The transaction starts at the handler and ends when it returns.

3/ will update soon in a new commit

heiwen: Understand. 

On (1) It's still easy to overlook the current scope, since it get's a bit mingled up with the actual method signature. Suggest to scope it as narrow as needed instead, just around the specific calls that use it.

On (2) where / how will the promise resolve?

packages/db/drizzle.ts
heiwen: Looked at the latest changes.

1/ You added type exports and build steps. Are they needed / consumed by anyone?
2/ Using wildcard syntax avoid exporting every file individually. This automatically adds files to the exports when they are added, reducing boilerplate. Given it's an internal package, there is no concern about over-exporting.

heiwen:
> > 3/ Right now the services don't correctly interfere the types from the models (during development). If helpful, this can be solved by making the services ElysiaJS instances:
> > See: "Best Practices - Do: Request dependent service as Elysia instance" (https://elysiajs.com/essential/best-practice.html#service)
> 
> The `any` isn’t coming from Elysia service composition; it’s caused by `withAudit` using `as any`, which erases Drizzle’s type inference, so `.returning()` becomes `any[]` and `[agent]` is inferred as `any`.
> 
> I can try to refactor `withAudit` to be fully typed, but it looks like overcomplicating the utility right now. What do you think?

Two inputs:
1/ The "any" that I mean is not coming from the withAudit. It's on the {Agent/Branches}Schema objects when I'm trying to use them as a first row in a service. 

2/ It's best practice to avoid "any". Suggest to spend 15 minutes and ask ChatGPT whether it can convert the with-audit into a typed version. If it works, great. If not, mark as FUTURE.

packages/db/schema/agents.ts
```diff
@@ -11,5 +11,20 @@ export const agents = pgTable(
...
     ...audit,
   },
-  (table) => [uniqueIndex("unique_slug").on(sql`LOWER(${table.slug})`)],
+  (table) => [
+    // Case-insensitive unique slug for non-deleted rows
+    uniqueIndex("unique_slug")
+      .on(sql`LOWER(${table.slug})`)
+      .where(isNull(table.deletedAt)),
+
+    // Speed up listing with soft-delete filter and ordering by createdAt
```
heiwen: The new indexes and related schema migrations seem like premature optimisations to me that unnecessarily increase complexity. We can introduce them when really needed. Likely performance issues will be in different places.

packages/db/schema/branches.ts
```diff
@@ -23,5 +30,15 @@ export const branches = pgTable(
     uniqueIndex("unique_slug_per_agent")
       .on(table.agentId, sql`LOWER(${table.slug})`)
       .where(isNull(table.deletedAt)),
+
+    // Speed up list queries: WHERE agentId = ? AND deletedAt IS NULL ORDER BY createdAt
+    index("idx_branches_agent_created_at_not_deleted")
```
heiwen: The new indexes and related schema migrations seem like premature optimisations to me that unnecessarily increase complexity. We can introduce them when really needed. Likely performance issues will be in different places.

apps/api/src/modules/agents/index.ts
turisanapo: I added another helper `runInRequestTransaction` in apps/api/src/utils/request-db.ts.

The helper is used at the service level to create a request-scoped transaction context, and the promise resolves or rejects when the transaction callback finishes by committing on resolve or rolling back on reject.

packages/db/schema/branches.ts
turisanapo: It is a good practice to create indices on where condition fields.

packages/db/schema/agents.ts
turisanapo: It is a good practice to create indices on where condition fields.

packages/db/drizzle.ts
turisanapo: You are right, I removed the type build step. Also, I'm now using wildcards for the export.

apps/api/src/modules/agents/index.ts
turisanapo: I added a comment as well that should make it clear when/how to use the helper.

PR #56, opened 2025-08-21
OpenAI-compatible AI Gateway using Vercel AI SDK

heiwen:
## Summary by CodeRabbit

- **New Features**
  - API Gateway with Swagger docs, root greeting, configurable port, v1 endpoints for chat completions (streaming & non‑streaming), embeddings (single & batch), and models (list & detail). Includes OpenAI‑style error responses, provider routing for chat vs embeddings, request logging, and CORS.

- **Chores**
  - Package manifest, scripts, and TypeScript config to support build/dev/lint/type‑check/start workflows.

- **Documentation**
  - README gateway entry and example environment file with placeholder variables.


apps/gateway/src/modules/completions.ts
```diff
@@ -0,0 +1,81 @@
...
+    };
+
+    if (stream) {
+      const result = await streamText({
+        model: groq(model),
+        messages: messages,
+        temperature,
+      });
+      return result.toTextStreamResponse();
+    }
```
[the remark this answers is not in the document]
heiwen: It should work with the existing method. Will validate later.

apps/gateway/src/modules/completions.ts
```diff
@@ -0,0 +1,81 @@
...
+        },
+      ],
+      usage: r.usage && {
+        prompt_tokens: r.usage.inputTokens ?? 0,
+        completion_tokens: r.usage.outputTokens ?? 0,
+        total_tokens:
+          r.usage.totalTokens ??
+          (r.usage.inputTokens ?? 0) + (r.usage.outputTokens ?? 0),
+      },
+    };
```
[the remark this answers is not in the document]
heiwen: Valid. Will implement later.

apps/gateway/src/modules/embeddings.ts
```diff
@@ -0,0 +1,55 @@
+import { embed, embedMany } from "ai";
+import { Elysia, t } from "elysia";
+import { createVoyage } from "voyage-ai-provider";
+
+const voyage = createVoyage({ apiKey: process.env.VOYAGE_API_KEY! });
+
```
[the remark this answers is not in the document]
heiwen: Valid. Will implement later, also for GROQ_API_KEY.

apps/gateway/src/modules/completions.ts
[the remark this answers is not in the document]
heiwen: Is this really required. Doesn't generateText / toTextStreamResponse already generate the same shape?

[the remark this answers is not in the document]
heiwen: But then again, aren't you just recreating the same shape here:

    } catch (e: any) {
      // Map common provider errors
      set.status = 400;
      return {
        error: {
          message: e?.message ?? "Request failed",
          type: "invalid_request_error",
          code: e?.code ?? "bad_request",
        },
      };

apps/gateway/src/modules/embeddings.ts
```diff
@@ -0,0 +1,55 @@
...
+          object: "embedding",
+          embedding: e,
+          index: i,
+        })),
+        model,
+        usage: usage && {
+          prompt_tokens: usage.tokens ?? 0,
+          total_tokens: usage.tokens ?? 0,
+        },
+      };
```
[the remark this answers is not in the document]
heiwen: You're wrong.

apps/gateway/src/modules/embeddings.ts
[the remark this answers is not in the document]
heiwen: Won't do, AI SDK throws on first usage.

apps/gateway/src/middleware/provider.ts
```diff
@@ -0,0 +1,37 @@
+import { createGroq } from "@ai-sdk/groq";
+import { Elysia } from "elysia";
+import { createVoyage } from "voyage-ai-provider";
+
+import type { LanguageModel, EmbeddingModel } from "ai";
+
+const groq = createGroq({ apiKey: process.env.GROQ_API_KEY! });
+const voyage = createVoyage({ apiKey: process.env.VOYAGE_API_KEY! });
+
```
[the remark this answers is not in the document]
heiwen: Won't do. Vercel AI SDK already throws.

apps/gateway/src/middleware/oai-errors.ts
```diff
@@ -0,0 +1,65 @@
...
+      const text = await res.text().catch(() => "");
+      return {
+        error: {
+          message: text || `Upstream error (${res.status})`,
+          type: res.status >= 500 ? "server_error" : "invalid_request_error",
+          param: undefined,
+          code: undefined,
+        },
+      };
+    }
```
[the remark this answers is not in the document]
heiwen: Not needed.

apps/gateway/src/modules/models.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  .get(
+    "/:id",
+    ({ params }) => {
+      const { id } = params;
+      return { id, object: "model", owned_by: "gateway" };
+    },
+    {
+      params: t.Object(t.String({ enum: [...SUPPORTED_MODELS] })),
+    },
```
turisanapo: Is it standard to just return `model` as `object` value?

apps/gateway/src/modules/completions.ts
```diff
@@ -0,0 +1,84 @@
...
+          temperature,
+        });
+        return result.toTextStreamResponse();
+      }
+
+      const result = await generateText({
+        model: chatModel,
+        messages: messages as ModelMessage[],
+        temperature,
+      });
```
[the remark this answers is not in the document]
heiwen: toTextStreamResponse does not support a getErrorMessage object. To my understanding it will still throw errors that are handled by the framework. Correct me if I'm wrong.

apps/gateway/src/middleware/provider.ts
```diff
@@ -0,0 +1,53 @@
...
+const isEmbedding = (id: string) => /^voyage-/i.test(id);
+
+// FUTURE support AWS Bedrock
+const pickChat = (id: string): LanguageModel => groq(id);
+
+const pickEmbedding = (id: string): EmbeddingModel<string> =>
+  voyage.textEmbeddingModel(id);
+
+const chatOrThrow = (id: string): LanguageModel => {
+  if (isEmbedding(id)) {
```
turisanapo, line 46: This is not necessary. Shouldn't it reject any model that is not supported by the `/chat/completions` with a consistent validation error message?

apps/gateway/src/modules/completions.ts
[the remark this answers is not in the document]
heiwen: Since we need OpenAI compatibility we will stick with toTextStreamResponse

apps/gateway/src/middleware/provider.ts
```diff
@@ -0,0 +1,53 @@
+import { createGroq } from "@ai-sdk/groq";
+import { Elysia } from "elysia";
+import { createVoyage } from "voyage-ai-provider";
+
+import type { LanguageModel, EmbeddingModel } from "ai";
+
+const groq = createGroq({ apiKey: process.env.GROQ_API_KEY! });
+const voyage = createVoyage({ apiKey: process.env.VOYAGE_API_KEY! });
+
+const isEmbedding = (id: string) => /^voyage-/i.test(id);
```
turisanapo, line 15: This cannot scale with more embedding models. We may consider iterating over the `supported-models.json` so that chat models are already separated from embedding models.

apps/gateway/src/middleware/provider.ts
```diff
@@ -0,0 +1,49 @@
...
+const chatOrThrow = (id: string): LanguageModel => {
+  if (isEmbedding(id)) {
+    throw badRequest(`Model "${id}" is an embedding model`);
+  }
+  return pickChat(id);
+};
+
+const embeddingOrThrow = (id: string): EmbeddingModel<string> => {
+  if (!isEmbedding(id)) {
+    throw badRequest(`Model "${id}" is a chat model`);
```
turisanapo, line 55: This is not necessary. Shouldn't it reject any model that is not supported by the `/embeddings` with a consistent validation error message?

apps/gateway/src/modules/models.ts
```diff
@@ -0,0 +1,34 @@
...
+  name: "models",
+  prefix: "/models",
+})
+
+  .get("/", () => {
+    return {
+      object: "list",
+      data: SUPPORTED_MODELS.map((id) => ({
+        id,
+        object: "model",
```
turisanapo: Is it expected to just return `model` as a value for `object`? I'm missing the sense for this endpoint.

apps/gateway/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "name": "@hebo/gateway",
+  "version": "0.0.1",
+  "private": true,
+  "sideEffects": false,
+  "engines": {
+    "bun": ">=1.2.19"
+  },
+  "type": "module",
+  "scripts": {
+    "build": "bun build src/index.ts --compile --minify-whitespace --minify-syntax --target=bun --outfile=server",
```
turisanapo, line 11: Why not build the types as well? Is there no consumer who needs them?

apps/gateway/src/index.ts
```diff
@@ -0,0 +1,38 @@
...
+      }),
+    )
+    .use(oaiErrors)
+    .get("/", () => "🐵 Hebo AI Gateway says hello!")
+    .group("/v1", (app) => app.use(completions).use(embeddings).use(models));
+
+if (import.meta.main) {
+  const app = createApp().listen(PORT);
+  console.log(
+    `🐵 Hebo Gateway is running at ${app.server?.hostname}:${app.server?.port}`,
```
turisanapo:
```suggestion
    `🐵 Hebo Gateway is running at ${app.server!.url}`,
```

apps/gateway/src/middleware/provider.ts
heiwen: This call throws an error if you try to call /chat/completions with an embeddings model (e.g. voyage-embeddings).

apps/gateway/src/middleware/provider.ts
heiwen: Yes, that can be implemented when supporting more models.

apps/gateway/src/middleware/provider.ts
heiwen: This call throws an error if you try to call /embeddings with a chat model (e.g. llama-4).

apps/gateway/package.json
heiwen: There is no consumer who needs them right now.

apps/gateway/src/index.ts
heiwen: Done

apps/gateway/src/middleware/provider.ts
turisanapo: I understand it. Shouldn't it throw for not supported chat models as well? (e.g. llama-2) And shouldn't the error message be consistent? Why have a special case for embedding models when the case here could be simplified to _any model not recognised as a /chat/completions model_?

apps/gateway/src/middleware/provider.ts
turisanapo: Shall we add a future then?

apps/gateway/src/middleware/provider.ts
turisanapo: I understand it. Shouldn't it throw for not supported embedding models as well? (e.g. voyage-1) And shouldn't the error message be consistent? Why have a special case for chat models when the case here could be simplified to _any model not recognised as an /embeddings model_?

apps/gateway/src/modules/models.ts
heiwen: Right now it's just a dummy. In the future it might support more:
https://vercel.com/docs/ai-gateway/openai-compat#retrieve-model

apps/gateway/src/modules/models.ts
heiwen: Yes: https://vercel.com/docs/ai-gateway/openai-compat#retrieve-model

apps/gateway/src/middleware/provider.ts
heiwen: Added

apps/gateway/src/middleware/provider.ts
heiwen: These are two different cases. Both are required. Just pushed code for the second.

apps/gateway/src/middleware/provider.ts
heiwen: These are two different cases. Both are required. Just pushed code for the second.

turisanapo (approved):

PR #59, opened 2025-08-22
optimize build for production

turisanapo:
## Summary by CodeRabbit

* **New Features**
  * API and Gateway now build and run from compiled server artifacts for faster, more deterministic startup.
  * Local DB loader is lazy/async in dev and emits a clear error if a dev-only DB is used in compiled builds.

* **Chores**
  * Unified build/output to dist/server; added type-check and standardized start/clean scripts.
  * Raised TS target to ES2022; removed separate declaration-only config.
  * Simplified package exports and trimmed a DB runtime dependency.

* **Packaging**
  * Bumped Elysia and Vite versions; updated dev tooling deps.


heiwen:
LGTM

PR #88, opened 2025-10-24
DX improvements

turisanapo:
## Summary by CodeRabbit

* **Chores**
  * Added new database migration and infrastructure management scripts (`db:migrate`, `db:reset`, `infra:up`, `infra:down`)
  * Removed legacy development scripts
  * Added Docker Compose configuration for PostgreSQL
  * Upgraded TypeScript to v5.9.2

* **Documentation**
  * Updated local development setup guide with modular, step-by-step workflow
  * Added environment variables and infrastructure management instructions


package.json
```diff
@@ -11,14 +11,14 @@
...
-    "dev": "turbo run dev --parallel",
+    "dev": "sh -c 'trap \"bun run postdev\" EXIT INT TERM; VITE_API_URL=http://localhost:3001 VITE_GATEWAY_URL=http://localhost:3002 turbo run dev --parallel'",
     "format": "prettier --write .",
     "format:staged": "prettier --write",
     "lint": "turbo run lint",
     "lint:deps": "check-dependency-version-consistency .",
     "lint:staged": "eslint --cache --fix",
     "postinstall": "lefthook install && bun run db generate",
-    "predev": "docker start hebo-postgres 2>/dev/null || docker run -d --name hebo-postgres -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=password -e POSTGRES_DB=local -p 5432:5432 -v \"$(pwd)/packages/database/.local-storage/postgres:/var/lib/postgresql/data\" postgres:17.5",
+    "predev": "(docker start hebo-postgres 2>/dev/null || docker run -d --name hebo-postgres -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=password -e POSTGRES_DB=local -p 5432:5432 -v \"$(pwd)/packages/database/.local-storage/postgres:/var/lib/postgresql/data\" postgres:17.5) && until docker exec hebo-postgres pg_isready -U postgres -d local -h localhost >/dev/null 2>&1; do sleep 0.5; done && bun run db migrate dev",
```
heiwen: Feels like a lot of magic happening here. Is that how other OSS projects do it?

apps/console/.env.example
```diff
@@ -1,6 +1,3 @@
-VITE_API_URL=your-api_url
```
heiwen, line 1: Let's keep around to document for people that want to override it

package.json
```diff
@@ -11,14 +11,14 @@
     "build": "turbo run build",
     "clean": "git clean -fdx -e '.env*'",
     "db": "bun run --filter @hebo/database db",
-    "dev": "turbo run dev --parallel",
+    "dev": "sh -c 'trap \"bun run postdev\" EXIT INT TERM; VITE_API_URL=http://localhost:3001 VITE_GATEWAY_URL=http://localhost:3002 turbo run dev --parallel'",
```
heiwen: Alternative suggestion to avoid duplicating URLs / ports in package.json:

1/ Detect isLocalDev based on whether proces.env.TURBO_HASH variable is set (and VITE_{API,GATEWAY}_URL are NOT set)
2/ Default ports in console/service.ts point to the same defaults as in API / Gateway
3/ Still allow manual override of URLs in console .env file via VITE_{API,GATEWAY}_URL

package.json
```diff
@@ -11,15 +11,18 @@
     "build": "turbo run build",
     "clean": "git clean -fdx -e '.env*'",
     "db": "bun run --filter @hebo/database db",
-    "dev": "turbo run dev --parallel",
+    "db:start": "docker start hebo-postgres 2>/dev/null || docker run -d --name hebo-postgres -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=password -e POSTGRES_DB=local -p 5432:5432 -v \"$(pwd)/packages/database/.local-storage/postgres:/var/lib/postgresql/data\" postgres:17.5",
+    "db:wait": "until docker exec hebo-postgres pg_isready -U postgres -d local -h localhost >/dev/null 2>&1; do sleep 0.5; done",
```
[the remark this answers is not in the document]
turisanapo: Not necessary at the moment since the db not starting would be an indication of something else major going on.

apps/console/app/lib/service.ts
```diff
@@ -8,7 +8,7 @@ import type { Api } from "~api";
 
 const url = isDevLocal
```
heiwen: Can't we point devLocal to 3001 as well? The only difference would be whether the service worker is running or not.

apps/console/app/routes/_shell/sidebar-playground.tsx
```diff
@@ -1,13 +1,7 @@
...
 import { kyFetch } from "~console/lib/service";
 
-const VITE_GATEWAY_URL = import.meta.env.VITE_GATEWAY_URL?.trim();
-
-if (!VITE_GATEWAY_URL) {
-  console.error(
-    "VITE_GATEWAY_URL environment variable is not set. Please configure it in your environment."
-  );
-}
+const VITE_GATEWAY_URL = import.meta.env.VITE_GATEWAY_URL?.trim() || "http://localhost:3002";
```
heiwen: URL's should not be hardcoded in component code. They belong in service.ts.

package.json
```diff
@@ -11,15 +11,18 @@
...
+    "db:stop": "docker stop hebo-postgres 2>/dev/null || true",
+    "db:migrate": "bun run db migrate dev",
+    "predev": "bun run db:start && bun run db:wait && bun run db:migrate",
```
heiwen: Bun doesn't execute predev. Needs documentation instead.

package.json
```diff
@@ -11,15 +11,18 @@
...
+    "dev": "sh -c 'trap \"bun run db:stop\" EXIT INT TERM; turbo run dev --parallel'",
```
heiwen: Is this how other OSS projects do it?

apps/console/app/lib/service.ts
turisanapo: What do you mean? I do not understand

package.json
turisanapo: What do you mean bun doesn't execute predev?

package.json
turisanapo: I couldn’t find a clear example of an app (not a database) that runs its DB with Docker run for local dev and really focuses on DX. Do you have any examples I can refer to?

package.json
heiwen: When you call "bun dev" it doesn't not execute "predev". It needs to be run manually. This is different from npm, which automatically runs "predev" when you call "dev".

package.json
heiwen: Langfuse?

ChatGPT: "Which OSS projects require a database and have package.json scripts for migrations and docker setup": https://chatgpt.com/share/69003611-3abc-8008-80d7-a4c286e301d6

apps/console/app/lib/service.ts
heiwen: Ignore, I was wrong

package.json
turisanapo: It does for me. It doesn't for you?

package.json
turisanapo: It is a separate and documented script rather than automatically stopping it when a user exits dev. I updated our repo accordingly.

docker-compose.yaml
```diff
@@ -0,0 +1,16 @@
+services:
+  postgres:
+    container_name: hebo-postgres
+    image: postgres:17.5
+    restart: unless-stopped
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: password
+      POSTGRES_DB: local
```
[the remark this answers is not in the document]
turisanapo: It is for development only. No need to create additional friction at onboarding time. The same pattern is not used for deployed environments.

docker-compose.yaml
heiwen, line 1: This is for dev. All the other Dockerfiles and indeed the whole infra folder is for production. Would make sense to semantically distinguish that by putting it somewhere else.

package.json
```diff
@@ -11,15 +11,17 @@
     "build": "turbo run build",
     "clean": "git clean -fdx -e '.env*'",
     "db": "bun run --filter @hebo/database db",
```
heiwen: Why do we have this for DB but not for the others? Suggest to be consistent and not have it at all. Unless there is a strong reason to have it.

package.json
```diff
@@ -11,15 +11,17 @@
...
+    "db:migrate": "bun run db migrate dev",
+    "db:reset": "bun run db migrate reset --force",
+    "db:start": "docker compose -f infra/docker/docker-compose.yml up -d postgres",
```
heiwen: db:start and db:stop is conceptually not only the db, but any development infrastructure, suggest to rename to make that clear.

PR #92, opened 2025-11-16
refactor include query parameter

turisanapo:
## Summary by CodeRabbit

* **Changes**
  * "branches" field in agent responses/payloads is now optional.
  * API endpoints document additional error responses (404 and 409) for create/get/update/delete.
  * Query parameter for including branches switched from expand=branches to branches=true.

* **Refactor**
  * Unified branch-include handling to a single query-based approach across endpoints and client routes.

### DIFF
--- a/apps/api/src/modules/agents.ts
+++ b/apps/api/src/modules/agents.ts
@@ -62,7 +62,7 @@ export const agentsModule = new Elysia({
         ...agentsInputCreate.properties,
         defaultModel: supportedModelsEnum,
       }),
-      response: { 201: agents },
+      response: { 201: agents, 409: t.String() },
     },
   )
   .get(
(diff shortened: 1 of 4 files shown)


apps/api/src/modules/agents.ts
```diff
@@ -13,14 +14,8 @@ import {
 
 const agents = t.Object({
```
heiwen: I believe you can simply use the generated agents type now.

turisanapo: `branches` field is not optional in the generated agents type.

heiwen: It's an array though, which can simply be empty?

turisanapo:
This solves HEB-168 as well.

PR #94, opened 2025-11-17
Enable tools for streaming endpoint

buibaoanh:
## Summary by CodeRabbit

* **Refactor**
  * Standardized completion responses so streaming and non-streaming outputs are consistently OpenAI-compatible, preserving finish reasons and usage details.

* **New Features**
  * Streaming now uses Server-Sent Events with keep-alive for more reliable live output.
  * Added incremental tool-call updates in streams so tool activity appears progressively.
  * Added a single-response OpenAI-compatible format for non-stream requests.

### DIFF
--- a/apps/gateway/src/utils/converters.ts
+++ b/apps/gateway/src/utils/converters.ts
@@ -15,7 +15,7 @@ import {
   type OpenAICompatibleMessage,
   type OpenAICompatibleTool,
   type OpenAICompatibleToolChoice,
-  type OpenAICompatibleToolCallDelta, // Added this line
+  type OpenAICompatibleToolCallDelta,
 } from "./openai-compatible-api-schemas";
 
 function convertToModelContent(content: OpenAICompatibleContentPart[]) {
--- a/apps/gateway/src/utils/converters.ts
+++ b/apps/gateway/src/utils/converters.ts
@@ -183,7 +183,6 @@ export const toOpenAICompatibleMessage = (
   }
 
   if (result.reasoningText) {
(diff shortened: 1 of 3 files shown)


buibaoanh:
image
image
image

apps/gateway/src/utils/converters.ts
```diff
@@ -223,3 +225,163 @@ export const toToolChoice = (
...
+            };
+            enqueue({
+              id: streamId,
+              object: "chat.completion.chunk",
+              created: creationTime,
+              model,
+              // eslint-disable-next-line unicorn/no-null
+              choices: [{ index: 0, delta, finish_reason: null }],
+            });
+            break;
```
turisanapo, line 259: Checking the examples in the Chat Completions docs, it seems that the first chunk must include `choices[0].delta.role = \"assistant\"`.

Since we only set role inside the `text-delta` branch, pure tool-call (gemini does this) responses never emit that chunk, so clients won’t see a role.

Shall we emit a dedicated opening chunk (or include role in the non-text branches) to stay spec-compatible? (Assuming my initial assumption based on chat completion examples is correct)

buibaoanh: The role actually is only assistant in any case according to https://github.com/vercel/ai/blob/ce4e273d77573872c0e4ed03202b1b63f85a63ea/packages/openai-compatible/src/chat/openai-compatible-chat-language-model.ts#L754 

If gemini has issues with this, this will be handled separately in PR: https://github.com/heboai/hebo/pull/97

turisanapo (approved):

PR #97, opened 2025-11-17
Buibaoanh/heb 164 gateway support reasoning config

buibaoanh:
## Summary by CodeRabbit

* **New Features**
  * Configurable "reasoning" on model calls (enable, effort levels, optional token limits, exclude).
* **New Models**
  * Added adapters for Claude Opus 4.5, Google Gemini variants, OpenAI GPT OSS variants, and Cohere Embed v4.0.
* **API Changes**
  * POST /chat/completions accepts an optional reasoning field.
* **Refactor**
  * Dynamic adapter system for model/provider selection and provider-specific option mapping.
* **Tests**
  * Expanded tests covering reasoning, model transforms, provider option mappings, and prompt transformations.


buibaoanh:
image

apps/gateway/src/utils/converters.ts
```diff
@@ -16,8 +17,118 @@ import {
...
+  },
+  // Groq
+  {
+    matches: (model) =>
+      model.provider === "groq" && model.modelId.startsWith("openai/gpt-oss"),
+    transform: (reasoning) => {
+      return {
+        // cannot disable reasoning due to ai sdk lacking include_reasoning option
+        groq: {
+          reasoningEffort: reasoning.effort,
```
heiwen: What about the other parameters?

apps/gateway/src/utils/openai-compatible-api-schemas.ts
```diff
@@ -101,6 +101,15 @@ export const OpenAICompatibleFinishReason = t.Union([
...
 ]);
 
+export const OpenAICompatibleReasoning = t.Object({
+  enabled: t.Boolean(),
+  max_tokens: t.Optional(t.Number()),
+  effort: t.Optional(
+    t.Union([t.Literal("low"), t.Literal("medium"), t.Literal("high")]),
+  ),
+  exclude: t.Optional(t.Boolean()),
+});
```
turisanapo, line 121: 1/ The name `OpenAICompatibleReasoning` since it implies that these paramenters are compatible with OpenAI models while OpenAI models only support `effort` and `summary` (which is missing here).

2/ What happens if a user uses `summary` as a reasoning option on hebo?

apps/gateway/src/middlewares/models/model.ts
```diff
@@ -0,0 +1,30 @@
...
+export interface ModelAdapter {
+  getModelType(): string;
+  getModality(): "chat" | "embedding";
+  getDisplayName(): string;
+  getRateLimit(): number;
+  transformConfigs(options: OpenAICompatibleOptions): Record<string, any>;
+}
+
+export abstract class ModelAdapterBase implements ModelAdapter {
+  abstract getModelType(): string;
```
turisanapo: Why is this method exposed? What is using it?

Models adapters are initialised using the factor, and the caller in order to call the factory must know the model type already.

packages/shared-data/models/gpt.ts
```diff
@@ -0,0 +1,56 @@
...
+export const GPT_OSS_120B = defineModel({
+  type: "openai/gpt-oss-120b",
+  displayName: "OpenAI GPT OSS 120B",
+  family: "gpt",
+  rateLimit: DEFAULT_RATE_LIMIT,
+  modality: "chat",
+  providers: [
+    {
+      bedrock: "openai.gpt-oss-120b-1:0",
+      groq: "openai/gpt-oss-120b",
```
turisanapo: Since each provider is now mapping the exact model id internally, would it make sense to remove the mapping here or is it used somewhere else?

packages/shared-data/models/gpt.ts
```diff
@@ -0,0 +1,56 @@
...
+export const GPT_OSS_20B = defineModel({
+  type: "openai/gpt-oss-20b",
+  displayName: "OpenAI GPT OSS 20B",
+  family: "gpt",
+  rateLimit: DEFAULT_RATE_LIMIT,
+  modality: "chat",
+  providers: [
+    {
+      bedrock: "openai.gpt-oss-20b-1:0",
+      groq: "openai/gpt-oss-20b",
```
turisanapo: Since each provider is now mapping the exact model id internally, would it make sense to remove the mapping here or is it used somewhere else?

packages/shared-data/models/cohere.ts
```diff
@@ -0,0 +1,14 @@
+import { DEFAULT_RATE_LIMIT, defineModel } from "./base";
+
+export const COHERE_EMBED_V4 = defineModel({
+  type: "cohere/embed-v4.0",
+  displayName: "Cohere Embed v4.0",
+  family: "cohere",
+  rateLimit: DEFAULT_RATE_LIMIT,
+  providers: [
+    {
+      cohere: "embed-v4.0",
```
turisanapo: Since each provider is now mapping the exact model id internally, would it make sense to remove the mapping here or is it used somewhere else?

packages/shared-data/models/gemini.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export const GEMINI_2_5_FLASH_PREVIEW = defineModel({
+  type: "google/gemini-2.5-flash-preview-09-2025",
+  displayName: "Gemini 2.5 Flash Preview (Sep 2025)",
+  family: "gemini",
+  rateLimit: DEFAULT_RATE_LIMIT,
+  modality: "chat",
+  providers: [
+    {
+      vertex: "gemini-2.5-flash-preview-09-2025",
```
turisanapo: Since each provider is now mapping the exact model id internally, would it make sense to remove the mapping here or is it used somewhere else?

packages/shared-data/models/gemini.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export const GEMINI_2_5_FLASH_LITE_PREVIEW = defineModel({
+  type: "google/gemini-2.5-flash-lite-preview-09-2025",
+  displayName: "Gemini 2.5 Flash Lite Preview (Sep 2025)",
+  family: "gemini",
+  rateLimit: DEFAULT_RATE_LIMIT,
+  modality: "chat",
+  providers: [
+    {
+      vertex: "gemini-2.5-flash-lite-preview-09-2025",
```
turisanapo: Since each provider is now mapping the exact model id internally, would it make sense to remove the mapping here or is it used somewhere else?

apps/gateway/src/middlewares/providers/bedrock.ts
```diff
@@ -20,8 +20,47 @@ export class BedrockProviderAdapter
   private config?: BedrockProviderConfig;
   private credentials?: BedrockCredentials;
 
-  constructor(modelName: string) {
-    super("bedrock", modelName);
+  // modelType to modelId
```
turisanapo: Unnecessary comment.

apps/gateway/src/middlewares/providers/cohere.ts
```diff
@@ -11,8 +11,17 @@ export class CohereProviderAdapter
 {
   private config?: ApiKeyProviderConfig;
 
-  constructor(modelName: string) {
-    super("cohere", modelName);
+  // modelType to modelId
```
turisanapo: Unnecessary comment.

apps/gateway/src/middlewares/providers/groq.ts
```diff
@@ -11,8 +11,18 @@ export class GroqProviderAdapter
 {
   private config?: ApiKeyProviderConfig;
 
-  constructor(modelName: string) {
-    super("groq", modelName);
+  // modelType to modelId
```
turisanapo: Unnecessary comment.

apps/gateway/src/middlewares/providers/index.ts
```diff
@@ -19,25 +18,24 @@ export class ProviderAdapterFactory {
...
+      BedrockProviderAdapter,
+      CohereProviderAdapter,
+      GroqProviderAdapter,
+      VertexProviderAdapter,
     ];
 
-    for (const providerSlug of providerSlugs) {
-      try {
+    for (const ProviderClass of ALL_PROVIDER_CLASSES) {
+      const tempInstance = new ProviderClass(modelType);
```
turisanapo: Now priority for providers is hardcoded by the list `ALL_PROVIDER_CLASSES` rather than beign delegated to some configuration. Is this intentional?

apps/gateway/src/middlewares/providers/vertex.ts
```diff
@@ -12,8 +12,26 @@ export class VertexProviderAdapter
 {
   private config?: VertexProviderConfig;
 
-  constructor(modelName: string) {
-    super("vertex", modelName);
+  // modelType to modelId
```
turisanapo: Unnecessary comment.

apps/gateway/src/modules/completions.ts
```diff
@@ -37,6 +39,7 @@ export const completions = new Elysia({
       const toolSet = toToolSet(tools);
       const modelMessages = toModelMessages(messages);
       const coreToolChoice = toToolChoice(toolChoice);
+      const providerOptions = (reasoning ? { reasoning } : {}) as any;
```
turisanapo: Why is `reasoning` renamed to `providerOptions`?

packages/shared-data/models/base.ts
```diff
@@ -0,0 +1,28 @@
...
+  [providerSlug: string]: string;
+}
+
+export interface ModelDefinition {
+  type: string;
+  displayName: string;
+  rateLimit: number;
+  providers: ModelProviderMap[];
+  modality: "chat" | "embedding";
+  family: string;
```
turisanapo: What is `family` used for?

packages/shared-data/models/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  COHERE_EMBED_V4,
+] as const;
+
+export const MODELS_BY_TYPE: Record<string, ModelDefinition> =
+  Object.fromEntries(SUPPORTED_MODELS.map((m) => [m.type, m]));
+
+/**
+ * Transform OpenAI-compatible reasoning parameters into model-specific configuration.
+ */
+export const getReasoningConfig = (
```
turisanapo: Where is this method used?

apps/gateway/src/utils/openai-compatible-api-schemas.ts
buibaoanh: Our gateway is following convention https://vercel.com/docs/ai-gateway/openai-compat#reasoning-configuration

The `summary` in this case could mean `reasoning_content`, as long as `exclude` setting is `false`, it should be available

apps/gateway/src/modules/completions.ts
buibaoanh: This is not a renaming, `reasoning` is just one of the options to send to provider. `providerOptions` is just a wrapper.

apps/api/src/modules/branches.ts
```diff
@@ -85,7 +84,7 @@ export const branchesModule = new Elysia({
     {
       body: t.Object({
         name: branchesInputUpdate.properties.name,
-        models: t.Optional(Models),
+        models: t.Optional(t.Array(t.Any())),
```
turisanapo: It is ok to drop the enum that comes from `supportedModels` but we still want models to be of type `Models`. You can move the type definition on api.

apps/api/src/modules/providers.ts
```diff
@@ -79,20 +78,20 @@ export const providersModule = new Elysia({
       const branches = await dbClient.branches.findMany();
 
       const affectedBranches = branches.filter((branch) => {
-        const models = branch.models as Models;
-        return models.some((model) =>
+        const models = branch.models as any[];
```
turisanapo: It is ok to drop the enum that comes from `supportedModels` but we still want models to be of type `Models`. You can move the type definition on api.

apps/api/src/modules/providers.ts
```diff
@@ -79,20 +78,20 @@ export const providersModule = new Elysia({
...
+        return models.some((model: any) =>
           model.routing?.only?.includes(params.slug),
         );
       });
 
       // Batch update all affected branches + delete provider in a single transaction
       await dbClient.$transaction([
         ...affectedBranches.map((branch) => {
-          const models = branch.models as Models;
+          const models = branch.models as any[];
```
turisanapo: It is ok to drop the enum that comes from `supportedModels` but we still want models to be of type `Models`. You can move the type definition on api.

apps/api/src/modules/providers.ts
```diff
@@ -79,20 +78,20 @@ export const providersModule = new Elysia({
...
           return dbClient.branches.update({
             where: { id: branch.id },
             data: {
-              models: models.map((model) =>
+              models: models.map((model: any) =>
```
turisanapo: It is ok to drop the enum that comes from `supportedModels` but we still want models to be of type `Models`. You can move the type definition on api.

apps/gateway/src/middlewares/model-config.ts
```diff
@@ -1,13 +1,22 @@
 import type { createDbClient } from "@hebo/database/client";
 import type { ProviderSlug } from "@hebo/database/src/types/providers";
-import type { Models } from "@hebo/shared-data/types/models";
+
+import { type SupportedModelType } from "./models";
+
+export interface ModelConfig {
```
turisanapo: If you define models on the API, you can just import from API rather than redifining it here.

apps/gateway/src/middlewares/model-config.ts
```diff
@@ -33,7 +42,7 @@ export class ModelConfigService {
       where: { agent_slug: agentSlug, slug: branchSlug },
       select: { models: true },
     });
-    const model = (branch.models as Models)?.find(
+    const model = (branch.models as unknown as ModelConfig[])?.find(
```
turisanapo: This shouldn't be necessary if you are just importing from a different folder.

apps/gateway/src/modules/models.ts
```diff
@@ -26,23 +27,28 @@ const model = t.Object({
 const modelsInclude = t.Optional(t.Object({ endpoints: t.Boolean() }));
 
 function modelToModelsResponse(
-  m: (typeof supportedModels)[0],
+  m: ReturnType<typeof ModelAdapterFactory.getAllModels>[0],
```
turisanapo: You can export Static of the underlying type and import it here.

buibaoanh:
Test results: 
image

apps/gateway/src/middlewares/models/model.ts
buibaoanh: this just plain object, exposing model properties is natural, I am not sure what is your concern here.

apps/gateway/src/middlewares/model-config.ts
```diff
@@ -33,7 +35,7 @@ export class ModelConfigService {
       where: { agent_slug: agentSlug, slug: branchSlug },
       select: { models: true },
     });
-    const model = (branch.models as Models)?.find(
+    const model = (branch.models as unknown as Models)?.find(
```
turisanapo: why is this necessary now?

packages/shared-api/types/model-config.ts
turisanapo, line 1: This must be defined using typebox as it was before.

apps/gateway/src/modules/models.ts
```diff
@@ -57,6 +60,7 @@ export const models = new Elysia({
   .get(
     "/",
     ({ query }) => {
+      const supportedModels = getSupportedModels();
```
heiwen: Nitpick: Inline getSupportedModels() since only used once

apps/gateway/src/modules/models.ts
```diff
@@ -77,7 +81,8 @@ export const models = new Elysia({
     "/:author/:slug/endpoints",
     ({ params }) => {
       const id = `${params.author}/${params.slug}`;
-      const m = supportedModels.find((model) => model.type === id);
+      const supportedModels = getSupportedModels();
```
heiwen: Nitpick: Inline getSupportedModels() since only used once

apps/gateway/src/modules/models.ts
```diff
@@ -25,24 +26,26 @@ const model = t.Object({
...
     },
     pricing: {
-      monthly_free_tokens: m.monthlyFreeTokens || 0,
+      monthly_free_tokens: m.pricing.monthly_free_tokens,
     },
     ...(withEndpoints && {
-      endpoints: Object.keys(m.providers[0]).map((tag) => ({ tag })),
+      endpoints: ProviderAdapterFactory.ALL_PROVIDER_CLASSES.filter(
+        (ProviderClass) => ProviderClass.supportsModel(m.id),
+      ).map((ProviderClass) => {
```
heiwen: Nitpick: Should ProviderClass more correctly be ProviderAdapterClass?

apps/gateway/src/middlewares/providers/index.ts
```diff
@@ -16,28 +15,26 @@ import { VertexProviderAdapter } from "./vertex";
 import type { ProviderAdapter } from "./provider";
 
 export class ProviderAdapterFactory {
+  static readonly ALL_PROVIDER_CLASSES = [
```
heiwen: Nitpick: Should this be ALL_PROVIDER_ADAPTER_CLASSES?

apps/gateway/src/middlewares/providers/provider.ts
```diff
@@ -2,33 +2,52 @@ import type {
...
   ) {}
 
-  getProviderModelId(): string {
-    const entry = supportedModels
-      .find((model) => model.type === this.modelName)
-      ?.providers.find((provider) => this.providerSlug in provider) as
-      | Record<ProviderSlug, string>
-      | undefined;
-    if (!entry) {
+  public getProviderSlug(): ProviderSlug {
```
heiwen: This feels like it can be a static method?

apps/gateway/src/middlewares/providers/provider.ts
```diff
@@ -2,33 +2,52 @@ import type {
...
 } from "@hebo/database/src/types/providers";
-import supportedModels from "@hebo/shared-data/json/supported-models";
 
 import type { Provider } from "ai";
 
 export interface ProviderAdapter {
   initialize(config?: ProviderConfig): Promise<this>;
   getProvider(): Promise<Provider>;
   resolveModelId(): Promise<string>;
+  transformConfigs(modelConfig?: Record<string, any>): Record<string, any>;
```
heiwen: IMHO ProviderAdapter shouldn't know about ModelConfig (which is a database type). Instead, it should just be passed the providerOptions.

apps/gateway/src/middlewares/providers/index.ts
```diff
@@ -16,28 +15,26 @@ import { VertexProviderAdapter } from "./vertex";
...
-          ),
-      ),
-    ];
-
-    for (const providerSlug of providerSlugs) {
-      try {
+    for (const ProviderClass of ProviderAdapterFactory.ALL_PROVIDER_CLASSES) {
+      if (ProviderClass.supportsModel(modelType)) {
+        const tempInstance = new ProviderClass(modelType);
+        const providerSlug = tempInstance.getProviderSlug();
```
heiwen: Looks like it should be static.

apps/gateway/src/middlewares/models/model.ts
```diff
@@ -0,0 +1,29 @@
+import type { OpenAICompatibleOptions } from "~gateway/utils/openai-compatible-api-schemas";
+
+export interface ModelAdapter {
+  getModelType(): string;
+  getModality(): "chat" | "embedding";
+  getDisplayName(): string;
+  getMonthlyFreeTokens(): number;
+  getOwner(): string;
+  getCreatedAt(): number;
+  transformConfigs(options?: OpenAICompatibleOptions): Record<string, unknown>;
```
heiwen: Let's be consistent in naming. Should we use "Options" or "Configs"?

apps/gateway/src/middlewares/models/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  owned_by: t.String(),
+  modality: t.Union([t.Literal("chat"), t.Literal("embedding")]),
+  pricing: t.Object({
+    monthly_free_tokens: t.Number(),
+  }),
+});
+
+export type SupportedModel = Static<typeof supportedModel>;
+
+const MODEL_ADAPTER_MAP = {
```
heiwen: Can you derive this from the Adapters itself, instead of maintaining a second map, same as you do with the Providers?

apps/gateway/src/middlewares/models/index.ts
```diff
@@ -0,0 +1,61 @@
...
+      throw new Error(`No model adapter found for model type: ${modelType}`);
+    }
+    return factoryMethod();
+  },
+};
+
+export function getSupportedModels(): SupportedModel[] {
+  return Object.values(MODEL_ADAPTER_MAP).map((factory) => {
+    const adapter = factory();
+    return {
```
heiwen: Any benefit of all these getters instead of the ModelAdapter returning an object of this shape?

apps/console/package.json
```diff
@@ -20,7 +20,7 @@
     "@conform-to/zod": "^1.13.3",
     "@elysiajs/eden": "^1.4.3",
     "@hebo/aikit-ui": "workspace:*",
-    "@hebo/shared-data": "workspace:*",
+
```
turisanapo:
```suggestion
```

apps/gateway/package.json
```diff
@@ -21,14 +21,14 @@
     "@ai-sdk/cohere": "^2.0.19",
     "@ai-sdk/google-vertex": "^3.0.51",
     "@ai-sdk/groq": "^2.0.28",
+    "@ai-sdk/provider-utils": "^3.0.19",
```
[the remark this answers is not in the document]
buibaoanh: it is latest version, I installed it with `bun add @ai-sdk/provider-utils`

apps/gateway/src/middlewares/providers/cohere.ts
```diff
@@ -11,8 +11,14 @@ export class CohereProviderAdapter
 {
   private config?: ApiKeyProviderConfig;
 
-  constructor(modelName: string) {
-    super("cohere", modelName);
+  static readonly providerSlug = "cohere";
+
+  static readonly SUPPORTED_MODELS_MAP: Record<string, string> = {
+    "cohere/embed-v4.0": "embed-v4.0",
+  };
```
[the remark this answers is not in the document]
buibaoanh: `supportsModel` is a static method inherited by all adapter classes from `ProviderAdapterBase`. It should work.

buibaoanh:
Re-testing:
```
  const { text, reasoningText, totalUsage } = await generateText({
    model: model,
    prompt: 'What is the capital of Germany?',
    providerOptions: {
      hebo: {
        reasoning: {
          enabled: true,
          effort: "low",
          excluded: true
        }
      }
    }
  });
```
image

**Excluding reasoning:**
image

**Console**
image

apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    return undefined;
+  }
+}
+
+export class Gemini25FlashPreviewAdapter extends GeminiModelAdapter {
+  readonly id = "google/gemini-2.5-flash-preview-09-2025";
+  readonly name = "Gemini 2.5 Flash Preview (Sep 2025)";
+  readonly owned_by = "google";
+  readonly created = 1_764_888_221;
```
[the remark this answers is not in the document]
buibaoanh: the created timestamps is `GMT: Thursday, December 4, 2025 10:43:41 PM`, it is not future date

turisanapo (approved):

apps/gateway/tests/middlewares/providers/provider.test.ts
```diff
@@ -0,0 +1,138 @@
...
+          safetySettings: [
+            {
+              category: "HARM_CATEGORY_HATE_SPEECH",
+              threshold: "BLOCK_LOW_AND_ABOVE",
+            },
+          ],
+        },
+      },
+    },
+  ];
```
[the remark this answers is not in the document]
buibaoanh: provider.test.ts only focus on testing transformation of provider name, not specific options of the model, you will see more coverage in model.test.ts

buibaoanh:
Retest:
image
image

apps/gateway/tests/middlewares/models/model.test.ts
```diff
@@ -0,0 +1,330 @@
...
+      },
+      expected: {
+        "openai-compatible": {
+          thinkingConfig: {
+            includeThoughts: false,
+            thinkingBudget: 8192,
+          },
+        },
+      },
+    },
```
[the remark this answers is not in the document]
buibaoanh: can you verify again, I cannot find any `excluded` in the code

apps/gateway/src/middlewares/models/cohere.ts
```diff
@@ -0,0 +1,25 @@
...
+  readonly modality = "embedding";
+  readonly name = "Cohere Embed v4.0";
+  readonly owned_by = "cohere";
+  readonly created = 1_764_888_221;
+  readonly pricing = {
+    monthly_free_tokens: 0,
+  };
+
+  transformOptions(options?: ProviderOptions): ProviderOptions {
+    const { openaiCompatible: openAiOptions, ...rest } = options || {};
```
heiwen: 1/ To check whether this works for both "OpenAI" and "OpenAICompatible" provider options?
2/ Since there is no actual transform happening, should there just be a default in the base class?

apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -0,0 +1,218 @@
...
+                },
+              };
+            }
+            return part;
+          }),
+        };
+      }
+      return message;
+    });
+  }
```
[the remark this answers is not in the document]
buibaoanh: it is an array, from github:
`export type LanguageModelV3Prompt = Array<LanguageModelV3Message>;`

apps/gateway/src/utils/converters.ts
```diff
@@ -105,12 +105,30 @@ export function toModelMessages(messages: OpenAICompatibleMessage[]) {
...
+              };
+
+              if (Object.keys(metadata).length > 0) {
+                content.providerOptions = {
+                  openaiCompatible: metadata,
+                };
+              }
+
+              return content;
+            }),
```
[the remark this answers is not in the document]
buibaoanh: I don't think this is needed, it is wrong to silence the error and worst if rethrow accidentally causing the misleading.
My PR has nothing to do with current logic of input.

apps/gateway/src/utils/converters.ts
```diff
@@ -170,14 +188,24 @@ export const toOpenAICompatibleMessage = (
...
+      return {
+        id: toolCall.toolCallId,
+        type: "function" as const,
+        function: {
+          name: toolCall.toolName,
+          arguments: JSON.stringify(toolCall.input),
+        },
+        ...metadata,
+      };
+    });
```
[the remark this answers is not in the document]
buibaoanh: This is unlikely to happen, I am following AI SDK implementation: https://raw.githubusercontent.com/vercel/ai/fad04b2e4ad6f927daebb5e7e342f0a98d35c3cd/packages/openai-compatible/src/chat/convert-to-openai-compatible-chat-messages.ts

buibaoanh:
Testing:
Receiving thought signature:
image
Pass through thought signature:
image
Inject dummy thought signature when not available:
image

apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -0,0 +1,218 @@
...
+        let hasInjected = false;
+        return {
+          ...message,
+          content: message.content.map((part: any) => {
+            if (part.type === "tool-call") {
+              if (hasInjected) {
+                return part;
+              }
+
+              const existingSignature = (part as any).providerOptions
```
heiwen: I don't think the signature is part of the providerOptions, is it? I thought it's part of the message parts instead?

apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -0,0 +1,218 @@
...
+              }
+
+              hasInjected = true;
+              return {
+                ...part,
+                providerOptions: {
+                  ...(part as any).providerOptions,
+                  google: {
+                    ...(part as any).providerOptions?.google,
+                    thoughtSignature: "context_engineering_is_the_way_to_go",
```
heiwen: Same as above comment.

apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -0,0 +1,218 @@
...
+  ): Record<string, any> {
+    const thinkingConfig: Record<string, any> = {};
+
+    switch (params.effort) {
+      case "minimal":
+      case "low": {
+        thinkingConfig.thinkingLevel = "low";
+        break;
+      }
+      case "medium": {
```
heiwen: Nitpick: collapse with high and xhigh?

apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -0,0 +1,218 @@
...
+    return undefined;
+  }
+
+  protected getThinkingConfig(
+    params: OpenAICompatibleReasoning,
+  ): Record<string, any> {
+    const thinkingConfig: Record<string, any> = {};
+
+    if (params.max_tokens === undefined) {
+      switch (params.effort) {
```
heiwen, line 68: What about "medium"?

apps/gateway/src/utils/converters.ts
```diff
@@ -105,12 +105,30 @@ export function toModelMessages(messages: OpenAICompatibleMessage[]) {
...
             role: "assistant",
-            content: message.tool_calls.map((toolCall) => ({
-              type: "tool-call",
-              toolCallId: toolCall.id,
-              toolName: toolCall.function.name,
-              input: JSON.parse(toolCall.function.arguments),
-            })),
+            content: message.tool_calls.map((toolCall) => {
+              const {
+                // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
heiwen, line 110: Which var is unused?

apps/gateway/src/utils/converters.ts
```diff
@@ -105,12 +105,30 @@ export function toModelMessages(messages: OpenAICompatibleMessage[]) {
...
+              } = toolCall;
+
+              const content: any = {
+                type: "tool-call",
+                toolCallId: id,
+                toolName: toolFunction.name,
+                input: JSON.parse(toolFunction.arguments),
+              };
+
+              if (Object.keys(metadata).length > 0) {
```
heiwen, line 124: Why are you moving the metadata from the content part into the providerOptions? Isn't the metadata part-specific and the provider options are call-level?

apps/gateway/src/utils/converters.ts
```diff
@@ -170,14 +188,24 @@ export const toOpenAICompatibleMessage = (
...
-      id: toolCall.toolCallId,
-      type: "function" as const,
-      function: {
-        name: toolCall.toolName,
-        arguments: JSON.stringify(toolCall.input),
-      },
-    }));
+    message.tool_calls = result.toolCalls.map((toolCall: any) => {
+      const metadata = {};
+      if (toolCall.providerMetadata) {
```
heiwen: Is providerMetadata the same as providerOptions above or is that different?

apps/gateway/src/middlewares/ai-model-factory.ts
```diff
@@ -29,31 +33,64 @@ export const aiModelFactory = new Elysia({
...
-        throw new BadRequestError(
-          `Model ${modelAliasPath} (${modelType}) is not supported.`,
-        );
-      }
-      if (modelModality !== modality)
-        throw new BadRequestError(
-          `Model ${modelAliasPath} (${modelType}) is not a ${modality} model.`,
+      const modelAdapter = ModelAdapterFactory.getAdapter(modelType);
+      if (modelAdapter.modality !== modality)
+        throw new Error(
```
turisanapo: When using `BadRequestError` you will return a 400 http status. If not, you will return a 500. I feel like here 400 is more apropriate.

apps/gateway/src/utils/openai-compatible-api-schemas.ts
```diff
@@ -122,4 +144,23 @@ export type OpenAICompatibleToolCallDelta = {
   index: number;
   type: "function";
   function: { name: string; arguments: string };
+  [key: string]: any;
 };
+
+export const OpenAICompatibleConfig = t.Object(
```
heiwen: Are OpenAICompatibleConfig and OpenAICompatibleOptions ever used?

apps/gateway/src/middlewares/models/gpt.ts
```diff
@@ -0,0 +1,74 @@
...
+      ...rest,
+      openaiCompatible: config,
+    };
+  }
+
+  private transformReasoning(
+    params: OpenAICompatibleReasoning,
+  ): Record<string, any> | undefined {
+    if (params.max_tokens !== undefined) {
+      throw new Error("GPT models do not support 'max_tokens' for reasoning.");
```
turisanapo: Shall we catch this and let the gateway return a 400? You can use `BadRequestError` upstream.

apps/gateway/src/middlewares/models/gpt.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+    if (params.exclude !== undefined) {
+      throw new Error("GPT models do not support 'exclude' for reasoning.");
```
turisanapo: Shall we catch this and let the gateway return a 400? You can use `BadRequestError` upstream.

apps/gateway/src/middlewares/models/index.ts
```diff
@@ -0,0 +1,40 @@
...
+for (const AdapterClass of ALL_MODEL_ADAPTER_CLASSES) {
+  const instance = new AdapterClass();
+  MODEL_ADAPTER_MAP[instance.id] = () => new AdapterClass();
+}
+
+export const ModelAdapterFactory = {
+  getAdapter(modelType: string): ModelAdapter {
+    const factoryMethod = MODEL_ADAPTER_MAP[modelType];
+    if (!factoryMethod) {
+      throw new Error(`No model adapter found for model type: ${modelType}`);
```
turisanapo: Shall we catch this and let the gateway return a 400? You can use `BadRequestError` upstream.

apps/gateway/src/middlewares/providers/index.ts
```diff
@@ -16,28 +15,27 @@ import { VertexProviderAdapter } from "./vertex";
...
+    for (const ProviderAdapterClass of ProviderAdapterFactory.ALL_PROVIDER_ADAPTER_CLASSES) {
+      if (ProviderAdapterClass.supportsModel(modelType)) {
+        return await this.createAdapter(
+          ProviderAdapterClass.providerSlug,
+          modelType,
+        );
       }
     }
+
     throw new Error(
```
turisanapo: Shall we catch this and let the gateway return a 400? You can use `BadRequestError` upstream.

apps/gateway/src/middlewares/providers/provider.ts
```diff
@@ -2,33 +2,60 @@ import type {
...
+  static supportsModel(modelType: string): boolean {
+    return !!this.getModelId(modelType);
+  }
+
+  async resolveModelId(): Promise<string> {
+    const modelId = (this.constructor as typeof ProviderAdapterBase).getModelId(
+      this.modelType,
+    );
+    if (!modelId) {
       throw new Error(
```
turisanapo: Shall we catch this and let the gateway return a 400? You can use `BadRequestError` upstream.

apps/gateway/src/middlewares/providers/index.ts
```diff
@@ -16,28 +15,27 @@ import { VertexProviderAdapter } from "./vertex";
...
-    for (const providerSlug of providerSlugs) {
-      try {
-        return await this.createAdapter(providerSlug, modelType);
-      } catch {
-        continue;
+    for (const ProviderAdapterClass of ProviderAdapterFactory.ALL_PROVIDER_ADAPTER_CLASSES) {
+      if (ProviderAdapterClass.supportsModel(modelType)) {
+        return await this.createAdapter(
+          ProviderAdapterClass.providerSlug,
+          modelType,
```
turisanapo, line 34: The provider fallback feature was relying on the exact order providers were defined in the supported model configuration Json. Now, instead we rely on the order of `ALL_PROVIDER_ADAPTER_CLASSES` which is the same for every model.

I feel like this new approach is error prone and we may end up with unexpected behaviours.

Is there a better way to frame this?

apps/gateway/src/middlewares/models/cohere.ts
buibaoanh: 1. The options is for OpenAICompatible, let me rename to avoid confusion
2. Yes, this can be moved to base class

apps/gateway/src/middlewares/models/gemini.ts
buibaoanh: yes, it is part of message (tool call), but in separate field,  and since this is provider-specific (google), we use `providerOptions` for injection.

apps/gateway/src/utils/converters.ts
buibaoanh: `type` is not being used, it is there for `metadata` extraction

apps/gateway/src/utils/converters.ts
buibaoanh: AI SDK does accept providerOptions at content-part level, which is not being mentioned in the documentation. 

FYR: https://github.com/vercel/ai/blob/main/packages/ai/src/prompt/content-part.ts#L82

apps/gateway/src/utils/converters.ts
buibaoanh: `providerOptions` is for input to generateText. However, for the results, `providerMetadata` represents the output.

apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -0,0 +1,220 @@
...
+
+              hasInjected = true;
+              return {
+                ...part,
+                providerOptions: {
+                  ...(part as any).providerOptions,
+                  google: {
+                    ...(part as any).providerOptions?.google,
+                    thoughtSignature: "context_engineering_is_the_way_to_go",
+                  },
```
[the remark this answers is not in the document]
buibaoanh: openaiCompatible is expected in transformPrompt method because input is generic, the purpose of transformation logic for adapter classes is to make it compatible to specific model / provider.

apps/gateway/src/utils/converters.ts
heiwen: What do you mean "metadata extraction"?

apps/gateway/tests/middlewares/providers/provider.test.ts
```diff
@@ -0,0 +1,154 @@
...
+      name: "Bedrock: no options returns empty object",
+      provider: bedrockProvider,
+      input: undefined,
+      expected: {},
+    },
+    {
+      name: "Bedrock: converts transformed openai-compatible options to snake_case",
+      provider: bedrockProvider,
+      input: {
+        openaiCompatible: {
```
heiwen: I'm not clear how these options are related to the requirements of this PR. Please explain why they are needed?

apps/gateway/tests/middlewares/providers/provider.test.ts
```diff
@@ -0,0 +1,154 @@
...
+          },
+        },
+      },
+    },
+    {
+      name: "Vertex: passes through generic safetySettings from openai-compatible",
+      provider: vertexProvider,
+      input: {
+        openaiCompatible: {
+          safetySettings: [
```
heiwen: These are not openAI compatible settings, they should be passed through via the google provider options instead.

apps/gateway/tests/middlewares/models/gemini.test.ts
heiwen, line 1: Please split into two files, one for reasoning, one for thought signatures.

apps/gateway/tests/middlewares/models/model.test.ts
```diff
@@ -0,0 +1,625 @@
...
+      inputPrompt: [
+        {
+          role: "assistant",
+          content: [
+            {
+              type: "tool-call",
+              toolCallId: "call_456",
+              toolName: "get_time",
+              args: { timezone: "UTC" },
+              providerOptions: {
```
heiwen: The official documentation uses a different format, see here:

https://ai.google.dev/gemini-api/docs/thought-signatures#openai

            "extra_content": {
              "google": {
                "thought_signature": "<Signature A>" //Required and Validated
              }
            },

apps/gateway/tests/middlewares/models/model.test.ts
```diff
@@ -0,0 +1,625 @@
...
+      ],
+      expectedPrompt: [
+        {
+          role: "assistant",
+          content: [
+            {
+              type: "tool-call",
+              toolCallId: "call_1",
+              toolName: "t1",
+              providerOptions: {
```
heiwen: See comment above.

apps/gateway/src/middlewares/models/gemini.ts
heiwen: This should be in extra_content, not in providerOptions according to official documentation:
https://ai.google.dev/gemini-api/docs/thought-signatures#openai

apps/gateway/src/utils/converters.ts
heiwen: So metadata = extra_content? Why not use the official nomenclature from the documentation?
https://ai.google.dev/gemini-api/docs/thought-signatures#openai

apps/gateway/src/utils/converters.ts
heiwen: Should this be extra_content as well?

apps/gateway/src/utils/converters.ts
```diff
@@ -316,13 +344,21 @@ export function toOpenAICompatibleStream(
           }
 
           case "tool-call": {
-            const { toolCallId, toolName, input } = part;
+            const { toolCallId, toolName, input, providerMetadata } = part;
+
+            const metadata = {};
+            if (providerMetadata) {
```
heiwen: Same as above, is this extra_content?

apps/gateway/tests/middlewares/models/gemini.test.ts
```diff
@@ -0,0 +1,625 @@
...
+      input: {
+        openaiCompatible: {
+          reasoning: {
+            enabled: true,
+          },
+        },
+      },
+      expected: {
+        openaiCompatible: {
+          thinkingConfig: {
```
heiwen, line 48: This is not openAiCompatible, it is google specific. Same for the rest of the Gemini section.

buibaoanh: The transformation of providerOptions happening in 2 steps in sequence:
1. Model adapter: convert options from OpenAI compatible into model specific options (effort, thinking budget, etc.)
2. Provider adapter: convert transformed options from step 1 to map to correct provider name, and special transformations required by that provider.

The test is about testing step 1, where we cannot tell what provider name is, as it is happening on provider adapter.

apps/gateway/src/utils/converters.ts
buibaoanh: providerMetadata is additional metadata (thought signature in this PR) passed from the provider to AI SDK

FYR: https://v6.ai-sdk.dev/docs/reference/ai-sdk-core/generate-text#total-usage.language-model-usage.reasoning-tokens

apps/gateway/tests/middlewares/models/model.test.ts
buibaoanh: Google provides native API (Hebo gateway using this) and OpenAI compatible API, they are different endpoints and expecting different input payload. On structure wise, extra_content is just another name of providerOptions.

In my PR, all I do is supply correct payload in LanguageModel format in Middleware, the AI SDK will take care of low level formatting for specific provider (google vertex). That's why you would not see extra_content anywhere in the code base.

Native payload:
```
{
            "functionCall": {
              "name": "check_flight",
              "args": {
                "flight": "AA100"
              }
            },
            "thoughtSignature": "<Signature A>" //Required and Validated
          }
```
OpenAI compatible payload:
```
{
            "extra_content": {
              "google": {
                "thought_signature": "<Signature A>" //Required and Validated
              }
            },
            "function": {
              "arguments": "{\"flight\":\"AA100\"}",
              "name": "check_flight"
            },
            "id": "function-call-1",
            "type": "function"
          }
```
The difference is the format, but structure is the same.

apps/gateway/tests/middlewares/models/gemini.test.ts
buibaoanh: sure, will split by model family as well.

apps/gateway/tests/middlewares/providers/provider.test.ts
buibaoanh: will update test to use more relevant options.

apps/gateway/src/middlewares/models/claude.ts
```diff
@@ -0,0 +1,107 @@
...
+        default: {
+          thinkingConfig.budgetTokens = Math.floor(MAX_CLAUDE_TOKENS * 0.5);
+          break;
+        }
+      }
+    } else {
+      thinkingConfig.budgetTokens = params.max_tokens;
+    }
+
+    if (thinkingConfig.budgetTokens < 1024) {
```
turisanapo, line 100: Does this mean that thinking is always enabled for Claude Opus 4.5?

apps/gateway/src/middlewares/providers/index.ts
```diff
@@ -16,28 +16,28 @@ import { VertexProviderAdapter } from "./vertex";
 import type { ProviderAdapter } from "./provider";
 
 export class ProviderAdapterFactory {
+  static readonly ALL_PROVIDER_ADAPTER_CLASSES = [
+    // FUTURE: error-prone for provider fallback feature
```
turisanapo:
```suggestion
    // FUTURE: error-prone for provider fallback feature. Fallback provider depends on the order of this list when the same model is supported by multiple providers.
```

apps/gateway/src/middlewares/models/claude.ts
buibaoanh: no, we have a check for `isReasoningActive` in the caller method, if false, we will not send this config.

apps/gateway/tests/middlewares/models/gemini.test.ts
```diff
@@ -0,0 +1,624 @@
...
+            includeThoughts: false,
+            thinkingLevel: "high",
+          },
+        },
+      },
```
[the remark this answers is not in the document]
buibaoanh: please relook, I have addressed this issue

apps/gateway/tests/middlewares/models/gemini.test.ts
```diff
@@ -0,0 +1,624 @@
...
+      expected: {
+        openaiCompatible: {
+          thinkingConfig: {
+            includeThoughts: true,
+            thinkingLevel: "medium",
+          },
+        },
+      },
+    },
```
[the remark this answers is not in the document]
buibaoanh: please relook, I have addressed this issue

PR #106, opened 2025-11-27
Turisanapo/heb 181 bug gateway doesnt throw authentication error when run

turisanapo:
## Release Notes

* **Bug Fixes**
  * Unified error handling across API and Gateway for consistent error responses
  * Improved streaming error detection and reporting
  * Enhanced authentication and validation error handling
  * Better database error mapping to HTTP status codes

* **New Features**
  * Standardized OpenAI-compatible error payload formatting
  * Client-side error state detection for improved reliability


turisanapo:
This solves [HEB-191](https://linear.app/heboai/issue/HEB-191/gateway-streaming-swallows-errors) as well.

apps/gateway/src/utils/converters.ts
```diff
@@ -239,113 +245,132 @@ export function toOpenAICompatibleStream(
...
-              id: streamId,
-              object: "chat.completion.chunk",
-              created: creationTime,
-              model,
-              // eslint-disable-next-line unicorn/no-null
-              choices: [{ index: 0, delta, finish_reason: null }],
-            });
-            break;
-          }
+      const enqueueError = (error: unknown) => {
```
buibaoanh, line 248: should we add an error log?

apps/gateway/src/middlewares/error-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+      "code" in error &&
+      error.code === "P2025"
+    )
+      return status(
+        404,
+        oaiError("Resource not found", "invalid_request_error", "not_found"),
+      );
+
+    return status(
+      500,
```
buibaoanh, line 67: should we add a detailed error log since this is server error?

apps/gateway/src/utils/converters.ts
turisanapo: Errors are already logged by the downstream error handler in `apps/gateway/src/middlewares/error-handler.ts`. It uses pino logger.

apps/gateway/src/middlewares/error-handler.ts
turisanapo: Errors are already logged by pino logger.

buibaoanh (approved):

apps/gateway/src/utils/converters.ts
```diff
@@ -239,113 +245,132 @@ export function toOpenAICompatibleStream(
...
-          case "error": {
-            console.error(
-              "[toOpenAICompatibleStream] Stream error:",
-              part.error,
-            );
-            controller.close();
-            return;
+            case "error": {
+              enqueueError(part.error);
+              return;
```
heiwen: Do you need another controller.close here?

apps/gateway/src/utils/converters.ts
```diff
@@ -239,113 +245,132 @@ export function toOpenAICompatibleStream(
...
           }
         }
-      }
 
-      controller.enqueue(encoder.encode("data: [DONE]\n\n"));
-      controller.close();
+        controller.enqueue(encoder.encode("data: [DONE]\n\n"));
+        controller.close();
+      } catch (error) {
+        enqueueError(error);
```
heiwen: Do you need another controller.close() here?

packages/shared-api/middlewares/auth/auth-service.ts
```diff
@@ -14,8 +15,8 @@ export const authService = new Elysia({ name: "auth-service" })
   .use(await createAuthService())
   .macro({
     isSignedIn: {
-      beforeHandle({ userId }) {
-        if (!userId) throw status(401, "Unauthorized");
+      resolve({ userId }) {
```
heiwen: Why the change from beforeHandle to resolve?

apps/gateway/src/middlewares/error-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+          error.code,
+        ),
+      );
+
+    // Prisma errors
+    if (
+      error &&
+      typeof error === "object" &&
+      "code" in error &&
+      error.code === "P2025"
```
heiwen: Suggest to have the prisma codes and error strings in a shared library, so you don't have to redefine it across different projects.

apps/gateway/src/middlewares/error-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+    const res = upstreamRes(error);
+    if (res) {
+      try {
+        return status(res.status, (await res.json()) as { error: unknown });
+      } catch {
+        const text = await res.text().catch(() => "");
+        return status(
+          res.status,
+          toOpenAiCompatibleError(
+            text || `Upstream error (${res.status})`,
```
heiwen: What does "upstream error" mean? Don't think it makes sense for our end users.

apps/gateway/src/middlewares/error-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+      return status(
+        error.status,
+        toOpenAiCompatibleError(
+          error.message,
+          "invalid_request_error",
+          error.code,
+        ),
+      );
+
+    // Prisma errors
```
heiwen: Suggest to keep the comments in the other blocks so we don't loose the knowledge of what type of error they are actually handling.

apps/gateway/src/middlewares/error-handler.ts
heiwen, line 1: Unclear what's actually new in this file vs. what's changed. Suggest to revert, rename oai-errors.ts to error-handlers.ts and then modify for a clear diff.

apps/gateway/src/utils/converters.ts
heiwen, line 1: Hard to review this change. What has actually changed vs what are only indentation changes. Any way to clarify that?

packages/aikit-ui/src/lib/openai-transport.ts
```diff
@@ -183,7 +189,9 @@ function handleSSEStream(
         ctrl.enqueue({ type: "start", messageId });
       },
       transform(result, ctrl) {
+        if (hasError) return;
```
heiwen, line 192: Is this hiding the actual error message?

packages/shared-api/middlewares/auth/stack-auth.ts
```diff
@@ -57,18 +51,16 @@ export const authServiceStackAuth = new Elysia({
...
         "Provide exactly one credential: Bearer API Key or JWT Header",
+        400,
       );
 
-    if (apiKey) return { userId: await checkApiKey(apiKey, log) } as const;
-    if (jwt) return { userId: await verifyJwt(jwt, log) } as const;
+    if (apiKey) return { userId: await checkApiKey(apiKey) } as const;
+    if (jwt) return { userId: await verifyJwt(jwt) } as const;
 
-    log.info("No credentials provided");
```
heiwen, line 71: This seems to be valuable logging information, why remove?

packages/shared-api/middlewares/auth/stack-auth.ts
```diff
@@ -46,7 +40,7 @@ const checkApiKey = async (
   if (res.status === 200) {
     const { user_id } = await res.json();
     return user_id;
-  } else log.info(res, "API Key check failed");
```
heiwen, line 49: This seems to be valuable logging information, why remove?

packages/shared-api/middlewares/auth/stack-auth.ts
```diff
@@ -14,22 +14,16 @@ const jwks = createRemoteJWKSet(
...
-const verifyJwt = async (
-  token: string,
-  log: Logger,
-): Promise<string | undefined> => {
+const verifyJwt = async (token: string): Promise<string | undefined> => {
   try {
     const { payload } = await jwtVerify(token, jwks);
     return payload.sub;
-  } catch (error) {
-    log.info({ err: error }, "JWT verification failed");
```
heiwen, line 25: This seems to be valuable logging information, why remove?

apps/gateway/src/utils/converters.ts
turisanapo: The `enqueueError` helper is already calling `controller.close()`.

apps/gateway/src/utils/converters.ts
turisanapo: The `enqueueError` helper is already calling `controller.close()`.

packages/shared-api/middlewares/auth/auth-service.ts
turisanapo: The `dbClient` factory is init using `.resolve()` and it needs the user id from auth when it is init.

By using `.resolve()` instead of `beforeHandle()` for the `isSignedIn` macro, we make sure that if no user id is available when the `dbClient` factory is called for the first time, we have thrown an authorization error.

apps/gateway/src/middlewares/error-handler.ts
turisanapo: Done.

apps/gateway/src/middlewares/error-handler.ts
turisanapo: Removed.

apps/gateway/src/middlewares/error-handler.ts
turisanapo: Now the code for prisma errors is self explicative. I removed this instance, but restored for other errors.

apps/gateway/src/middlewares/error-handler.ts
turisanapo: You can check this [commit](https://github.com/8monkey-ai/hebo/pull/106/commits/8185b48b49a5ac24a1548bc37ad59789da9cf11b)

apps/gateway/src/utils/converters.ts
turisanapo: In this change, I just wrapped part of the existing logic in a `try`/`catch` block. The actual changes are related to the `"error"` case and the `catch` block. I didn't change anything within the `try` (except for the error case).

packages/aikit-ui/src/lib/openai-transport.ts
turisanapo: Before any error `hasError` is false and set to true only once an error happens. This condition has been included to make sure that any residual streaming chunks are not returned to the client in case an error has happened.

apps/gateway/src/middlewares/error-handler.ts
```diff
@@ -0,0 +1,83 @@
...
+    const prismaError = getPrismaError(error);
+    if (prismaError)
+      return status(
+        prismaError.status,
+        toOpenAiCompatibleError(
+          prismaError.message,
+          "invalid_request_error",
+          "not_found",
+        ),
+      );
```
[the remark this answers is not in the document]
turisanapo: The gateway is creating no resource, just reading, so conflict can't happen.

apps/gateway/src/utils/converters.ts
heiwen: Ideally we scope try / catch to as narrow as possible. Which part of the code can actually throw?

packages/aikit-ui/src/lib/openai-transport.ts
heiwen: Why shouldn't we return residual streaming chunks?

packages/shared-api/middlewares/auth/auth-service.ts
heiwen: According to ElysiaJS documentation, authentication should be handled in beforeHandle. Why not throw in beforeHandle directly (like it was before). That way we never get to the DB Factory itself...?

packages/database/src/errors.ts
```diff
@@ -0,0 +1,12 @@
+const prismaErrorsHttpStatusMapping = {
+  P2002: { status: 409, message: "Resource already exists" },
+  P2025: { status: 404, message: "Resource not found" },
+} as const;
+
+export const getPrismaError = (error: unknown) =>
```
heiwen: rename to identifyPrismaError

apps/gateway/src/utils/converters.ts
turisanapo: The scope of try / catch is correct. The part that throws is `for await (const part of result.fullStream)`. Basically, each time it awaits the next chunk from the stream.

packages/aikit-ui/src/lib/openai-transport.ts
turisanapo: If you return residual streaming chunks you may update the ui showing the message content _after_ the user has already received an error, making the feedback confusing.

Here I want to make sure the client receives a consistent stream: either successful content OR an error, never a confusing mix.

packages/database/src/errors.ts
turisanapo: Done!

packages/shared-api/middlewares/auth/stack-auth.ts
turisanapo: Replaced with error message for both us and the user.

packages/shared-api/middlewares/auth/stack-auth.ts
turisanapo: Replaced with error message for both us and the user.

packages/shared-api/middlewares/auth/stack-auth.ts
turisanapo: Replaced with error message for both us and the user.

packages/shared-api/middlewares/auth/auth-service.ts
turisanapo: Restored.

packages/shared-api/middlewares/auth/stack-auth.ts
```diff
@@ -59,11 +59,10 @@ export const authServiceStackAuth = new Elysia({
...
 
-    if (apiKey && jwt)
-      throw status(
-        400,
-        "Provide exactly one credential: Bearer API Key or JWT Header",
-      );
+    if (apiKey && jwt) {
+      log.info("User provided both API Key and JWT Header");
+      return { userId: undefined } as const;
+    }
```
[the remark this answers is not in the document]
turisanapo: We don't want the auth service to raise errors.

PR #136, opened 2025-12-24
Buibaoanh/heb 164 split gemini reasoning basic

buibaoanh:
## Summary by CodeRabbit

* **New Features**
  * Added configurable reasoning/thinking support for Gemini and GPT models with effort levels: none, minimal, low, medium, high, xhigh.
  * Completions now accept `reasoning` and `reasoning_effort` to control model reasoning behavior.

* **API Changes**
  * Completions endpoint field renamed: `toolChoice` → `tool_choice`.

* **Tests**
  * Added test suites validating reasoning option transformations across models and providers.


apps/gateway/tests/middlewares/providers/provider.test.ts
heiwen, line 1: Split by provider, same as the models

apps/gateway/tests/middlewares/models/gemini.thought-signature.test.ts
heiwen, line 1: Thought signatures should not be part of this PR.

apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -1,14 +1,96 @@
...
 
 export abstract class GeminiModelAdapter extends ModelAdapterBase {
   readonly modality = "chat";
   readonly owned_by = "google";
   readonly pricing = {
     monthly_free_tokens: 0,
   };
+
+  transformOptions(options?: ProviderOptions): ProviderOptions {
+    const modelConfig: Record<string, any> = {};
```
heiwen: Nitpick: any specific reason we use "config" for model vs "options" for provider? If no specific reason, let's streamline. Suggest Options, as modelConfig has overlapping meaning in the same repo.

(same comment applies to gpt.ts)

apps/gateway/src/modules/completions.ts
```diff
@@ -27,15 +28,21 @@ export const completions = new Elysia({
         model: modelAliasPath,
         messages,
         tools,
-        toolChoice,
+        tool_choice,
+        reasoning,
```
heiwen, line 35: The original version of the OpenAI chat/completions API only offered the basic `reasoning_effort` parameter. Newer ones introduced the more powerful `reasoning: { reasoning_effort: ..., ... }` object. To test whether the OpenAI SDK, Langchain SDK & Vercel AI SDK work if we just support the `reasoning` object in our API, or we also have to support the `reasoning_effort` parameter for backwards compatibility.

apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -1,14 +1,96 @@
...
+
+    if (options?.openaiCompatible) {
+      const reasoning = (options.openaiCompatible as any)
```
heiwen: The classes need to know external conventions like `openAiCompatible` and `modelConfig` which leads to risk in implementation mistakes. Instead, inject just the sub-object into transformOptions and handle naming conventions / object hierachy at the caller side. Also reduces boilerplate in the implementation code of transformOptions hooks.

(Same applies to transformOptions in the providers)

apps/gateway/src/middlewares/providers/vertex.ts
```diff
@@ -24,6 +25,21 @@ export class VertexProviderAdapter
     super(modelType);
   }
 
+  transformOptions(options?: ProviderOptions): ProviderOptions {
```
heiwen: This looks exactly the same as the base class, can it be removed from here?

apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -1,14 +1,96 @@
...
+import type { ProviderOptions } from "@ai-sdk/provider-utils";
 
 export abstract class GeminiModelAdapter extends ModelAdapterBase {
   readonly modality = "chat";
   readonly owned_by = "google";
   readonly pricing = {
     monthly_free_tokens: 0,
   };
+
+  transformOptions(options?: ProviderOptions): ProviderOptions {
```
heiwen: 1/ The `transformOptions` methods need to know external conventions like `openAiCompatible` and `modelConfig`. This leads to risk in implementation mistakes. Instead, suggest to directly inject just the sub-object into `transformOptions` and handle naming conventions / object hierachy at the caller side. That also reduces boilerplate in the implementation code of `transformOptions` hooks.

2/ Any specific reason we call it `modelConfig` but `providerOptions`? If no specific reason, let's streamline. Suggest `modelOptions`, as `modelConfig` has overlapping meaning in the same repo. Maybe even better, use `hebo` / `heboInternal`. Or, maybe even simpler, we just remove that layer from then object hierachy. This would also apply to the original object put into the `providerOptions` in `completions.ts` (as we never actually use an openaiCompatible _provider_ itself).

3/ Do I understand correctly that you are returning the original options PLUS the transformed ones? Is there any reason we need the original ones? If not, it creates confusion and suggest we only return the transformed ones. 

(same comment applies to gpt.ts and bedrock/vertex.ts)

apps/gateway/src/middlewares/providers/bedrock.ts
```diff
@@ -33,6 +34,50 @@ export class BedrockProviderAdapter
...
+    let modifiedOptions: ProviderOptions = { ...rest };
+
+    if (modelConfig) {
+      const snakeCaseConfig =
+        BedrockProviderAdapter.convertObjectKeysToSnakeCase(
+          modelConfig as Record<string, any>,
+        );
+      modifiedOptions = {
+        ...modifiedOptions,
+        bedrock: {
```
heiwen: Related to my comments on the transformOptions hook in models, I see two options here:

1/ We might need to infer the term "bedrock" from the external caller. This could be done from the slug, or, since sometimes the slug and the key for the provider options might vary, introduce an additional field on the `ProviderAdapter` like `providerOptionsKey` that can be used from the aiModelFactory.

2/ If the intermediary representation of the `providerOptions` doesn't have any key like `openaiCompatible` or `hebo` / `heboInternal` or the like, then this code can continue to work as is (but probably also removing the `modifiedOptions`, that I don't believe we need).

apps/gateway/src/middlewares/models/gemini.ts
buibaoanh: 1/ and 2/ Make sense, object hierarchy is the cause of the ugly part. Original intention was skipping the "openaiCompatible" / "internal" layer by passing only the options. But this got changed later due to ambiguity of maintaining same data structure across adapters, which was meant to create flexibility and consistency. Now look like it is creating confusion / inconsistency rather than to be useful. I will update accordingly to keep options simpler.

3/ This happens due to 1/ and 2/ and the assumption that different adapter may expect to transform based on original options, but again, this is confusing and against polymorphism. Thanks for pointing out, I will adjust accordingly.

apps/gateway/src/middlewares/providers/bedrock.ts
buibaoanh: Make sense, let me do that.

turisanapo (approved):

PR #137, opened 2025-12-24
Buibaoanh/heb 164 split gemini reasoning thought signature

buibaoanh:
This reverts commit 321c6cdb881228f35895538ab9cadda05d532ef8.


apps/gateway/package.json
```diff
@@ -22,6 +22,7 @@
     "@ai-sdk/cohere": "^2.0.19",
     "@ai-sdk/google-vertex": "^3.0.51",
     "@ai-sdk/groq": "^2.0.28",
+    "@ai-sdk/provider": "^3.0.0",
```
heiwen, line 25: Version 3 of @ai-sdk/provider (and with that LanguageModelV3...) was published with AI-SDK v6. 

Since you are requiring this, you also need to
1/ Bump version of ai to ^6
2/ Check whether there is anything else we need to migrate:

https://vercel.com/blog/ai-sdk-6

apps/gateway/src/utils/openai-compatible-api-schemas.ts
```diff
@@ -31,6 +31,7 @@ export const OpenAICompatibleMessageToolCall = t.Object({
     arguments: t.String(),
     name: t.String(),
   }),
+  extra_content: t.Optional(t.Object({}, { additionalProperties: true })),
```
heiwen, line 34: Since we are supporting extra_content, it makes sense to also support the other extra_body parameters at the same time. Otherwise the API might be confusing to some people.

https://ai.google.dev/gemini-api/docs/openai

apps/gateway/src/utils/converters.ts
```diff
@@ -332,13 +335,14 @@ export function toOpenAICompatibleStream(
           }
 
           case "tool-call": {
-            const { toolCallId, toolName, input } = part;
+            const { toolCallId, toolName, input, providerMetadata } = part;
```
heiwen, line 338: I still confused about the term `providerMetadata` here. Shouldn't this be `providerOptions` according to `LanguageModelV2ToolCallPart`: https://ai-sdk.dev/providers/community-providers/custom-providers#tool-call-parts

apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -86,7 +87,40 @@ export abstract class GeminiModelAdapter extends ModelAdapterBase {
   }
 }
 
-export abstract class Gemini3ModelAdapter extends GeminiModelAdapter {}
+export abstract class Gemini3ModelAdapter extends GeminiModelAdapter {
+  transformPrompt(prompt: LanguageModelV3Prompt): LanguageModelV3Prompt {
```
heiwen, line 91: Is this code correctly handling the case of sequential vs parallel function calls? As outlined here:

https://ai.google.dev/gemini-api/docs/thought-signatures#function-calling

apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -86,7 +87,40 @@ export abstract class GeminiModelAdapter extends ModelAdapterBase {
...
+      if (message.role === "assistant" && Array.isArray(message.content)) {
+        let hasInjected = false;
+        return {
+          ...message,
+          content: message.content.map((part) => {
+            if (part.type === "tool-call") {
+              if (hasInjected) {
+                return part;
+              }
+              const thoughtSignature =
```
heiwen, line 102: 1/ If the thoughtSignature exists, we should not modify anything, just pass the original object through.
2/ If we inject a thoughtSignature, we should log this so we can later show this in the console, since it can lead to suboptimal model performance.
3/ To confirm that if no provider specific behavior is implemented, passthrough is the general behavior for extra_content (meaning it is not stripped along the way)

apps/gateway/package.json
buibaoanh: The reason for adding `@ai-sdk/provider` is about reusing `LanguageModelV3Prompt` for type hint in typescript, I could just lower it to `LanguageModelV2Prompt` (available in current package too) to limit scope of this PR rather than upgrade AI SDK to v6 (which makes better sense to be done in separate PR).

apps/gateway/src/utils/openai-compatible-api-schemas.ts
buibaoanh: For `cached_content`:
- Is this same as https://platform.openai.com/docs/guides/prompt-caching in OpenAI, should it be named as `prompt_cache_key` for better clarity instead?
- This feature seems to be out of scope, does it make sense to implement it in a separate PR?

For `thinking_config`:
- I find this is what we have done for basic reasoning, assuming it is already supported.

apps/gateway/src/utils/converters.ts
buibaoanh: Understand your confusion.

In AI SDK, they have different interfaces for input and output:
- `LanguageModelV3Prompt`: The specification for what you send to the model (input)
- `LanguageModelV3Content`: The specification for what the models generate (output)

`providerOptions` and `providerMetadata` have same data structure but for different usage, one is for input, another is for output.

`LanguageModelV3ToolCallPart` is part of `LanguageModelV3Prompt`, which supports `providerOptions`.

In my PR, the `part` here is actually an instance of `LanguageModelV3ToolCall` which is part of  `LanguageModelV3Content`, which supports providerMetadata (the docs is out of sync with the [code](https://github.com/vercel/ai/blob/7045f986afe2d9ad1ef09ee8d9596d1ac00e88ab/packages/provider/src/language-model/v2/language-model-v2-tool-call.ts)

Another place to check `providerMetadata` definition is [generateText > returns](https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-text#returns)

apps/gateway/src/middlewares/models/gemini.ts
buibaoanh: Yes, parallel function calls have more than one tool call, `hasInjected` is there to make sure the signature injection is only done for the first tool call. 

You may refer to unit test `tests/middlewares/models/gemini.thought-signature.test.ts`.

apps/gateway/src/middlewares/models/gemini.ts
buibaoanh: Good catch!

My implementation will accidentally strip away other options in extra_content object. I can correct this with pass through logic.

For 2/, I don't understand how logging is being implemented, is this a good example? [Transport MCP](https://github.com/8monkey-ai/hebo/blob/687023ec07420c3954725c15d2a13de935414fdb/apps/mcp/src/aikit/mcp-transport.ts#L9)

apps/gateway/package.json
heiwen: According to the release notes, AI SDK 6 has minimal breaking changes. Suggest to spend 10-15 minutes checking whether there is anything breaking / deprecated that we need to migrate. If minimal, do it as part of this PR. Otherwise depend on v2 of @ai-sdk/provider.

apps/gateway/src/middlewares/models/gemini.ts
heiwen: In the documentations there are examples that have multiple signatures (`<Signature A> and <Signature B>`)). Is that correctly handled?

apps/gateway/src/middlewares/models/gemini.ts
heiwen: By default, all extra_content should be passed through two ways. This should ideally happen on the generic openai converter layer, not on the provider-specific layer, so that we don't have to re-implement it for every individual provider. 

For logging, createGateway has a shared PinoLogging instance that is part of the ElysiaJS context. Need to identify what's the best pattern to access / reuse that without recreating. Or just recreate it, if cheap. Related to that, there is warnings[] object part of the streamText, maybe we can fill that within the provider and then do one logging call in the /chat/completions itself.

apps/gateway/src/utils/converters.ts
heiwen: Clear, tq

apps/gateway/src/utils/openai-compatible-api-schemas.ts
heiwen: I'm not talking about cached_content, I'm talking about extra_body. Extra_body (like extra_content) is a catch-all which we want to handle on the openAICompatible layer. It should automatically attached to providerOptions if present, no matter what's in there. We don't have to know about the details of the shape. The providerMetadata on the other side should then automatically convert back into extra_xzy.

See here how it can be used: https://ai.google.dev/gemini-api/docs/openai

apps/gateway/src/middlewares/models/gemini.ts
buibaoanh: Yes, current injection logic will maintain signature A and signature B too.

PR #138, opened 2025-12-24
Buibaoanh/heb 164 split claude reasoning support

buibaoanh:


buibaoanh:
This will be implemented in Hebo Gateway.

PR #140, opened 2026-01-01
Buibaoanh/heb 164 split gemini reasoning thought signature

buibaoanh:
## Summary by CodeRabbit

* **New Features**
  * Richer provider integration: metadata now flows through requests, results, and live streams; per-request provider options supported; post-generation and streaming transformation hooks added.
  * New utilities to normalize metadata casing between providers (camelCase ↔ snake_case).

* **Bug Fixes**
  * Improved type safety and more consistent prompt/result handling across adapters.

* **Tests**
  * Added tests validating metadata/options propagation and provider transformation behavior.


apps/gateway/src/middlewares/models/gemini.ts
```diff
@@ -86,7 +87,48 @@ export abstract class GeminiModelAdapter extends ModelAdapterBase {
...
+      if (message.role === "assistant" && Array.isArray(message.content)) {
+        let hasInjected = false;
+        return {
+          ...message,
+          content: message.content.map((part) => {
+            if (part.type === "tool-call") {
+              if (hasInjected) {
+                return part;
+              }
+              this.logger?.warn(
```
heiwen: 1/ I don't think this code is correct. If I read correctly, it also issues a warning when it passes through the original thought signature (compared to only issuing the warning when injecting the dummy signature). Suggest to refactor this method, so that it returns the original object without modifications if there is already a thought signature in place.

2/ To avoid passing the logger instance through to the deep level, suggest to use the built-in "warnings" property in AI SDK and then use the log object in /chat/completions itself to log any warning from the {generate/stream}Text results.

apps/gateway/src/utils/openai-compatible-api-schemas.ts
```diff
@@ -24,14 +24,22 @@ export const OpenAICompatibleContentPartFile = t.Object({
...
+    id: t.String(),
+    function: t.Object({
+      arguments: t.String(),
+      name: t.String(),
+    }),
+  },
+  {
+    additionalProperties: false,
+    patternProperties: {
+      "^extra_": t.Optional(t.Object({}, { additionalProperties: true })),
```
heiwen: I think you are misunderstanding the usage of extra_body. See my comment further down.

apps/gateway/src/utils/openai-compatible-api-schemas.ts
```diff
@@ -52,13 +60,21 @@ export const OpenAICompatibleUserMessage = t.Object({
...
+    role: t.Literal("assistant"),
+    content: t.Union([t.String(), t.Null()]),
+    tool_calls: t.Optional(t.Array(OpenAICompatibleMessageToolCall)),
+    reasoning: t.Optional(t.String()),
+    reasoning_content: t.Optional(t.String()),
+  },
+  {
+    additionalProperties: false,
+    patternProperties: {
+      "^extra_": t.Optional(t.Object({}, { additionalProperties: true })),
```
heiwen: I think you are misunderstanding the usage of extra_body. See my comment further down.

apps/gateway/tests/utils/converters.test.ts
heiwen, line 1: Suggest to add test case that tests whether other things in extra_body (e.g. cached_content) are properly passed through.

apps/gateway/package.json
```diff
@@ -22,6 +22,7 @@
     "@ai-sdk/cohere": "^2.0.19",
     "@ai-sdk/google-vertex": "^3.0.51",
     "@ai-sdk/groq": "^2.0.28",
+    "@ai-sdk/provider": "^3.0.0",
```
heiwen: Let's downgrade to ^2 for this PR and move to v6 of AI SDK in a separate one.

apps/gateway/tests/utils/converters.test.ts
```diff
@@ -228,4 +230,87 @@ describe("toModelMessages", () => {
...
+
+  describe("merges extra_ properties into providerOptions", () => {
+    const testCases = [
+      {
+        name: "should merge extra_ properties for assistant message without tool calls",
+        inputMessages: [
+          {
+            role: "assistant",
+            content: "Hello",
+            extra_body: { reasoning_effort: "high" },
```
heiwen: I think you are misunderstanding the purpose of extra_body. It doesn't exist on the message object to my understanding. It doesn't really exist at all in fact. It's a helper on the SDK-level by OpenAI to push parameter directly into the body of the request.

What we need to do is identify all parameters in the body that are not part of the OpenAICompatible schema and push them into providerOptions.

Now Google also misunderstood that, so what they did is to put an extra_body object into an extra_body object, like here: https://ai.google.dev/gemini-api/docs/openai#extra-body

So we should have some logic which handles this special case and simply unpacks that into providerOptions (without the additional extra_body).

apps/gateway/src/utils/converters.ts
```diff
@@ -332,13 +352,14 @@ export function toOpenAICompatibleStream(
           }
 
           case "tool-call": {
-            const { toolCallId, toolName, input } = part;
+            const { toolCallId, toolName, input, providerMetadata } = part;
```
heiwen, line 366: To double check whether we need to support extra_part in any way:
https://docs.cloud.google.com/vertex-ai/generative-ai/docs/migrate/openai/overview#extra_part_features

apps/gateway/src/middlewares/models/gemini.ts
buibaoanh: 1/ Fixed. 

2/ "warnings" is only available on response side, transformPrompt logic applied on request side, and it is stateless. To achieve this, we probably need to do it at doGenerate and doStream level, which is more complex than current logging, what do you think?

apps/gateway/src/utils/converters.ts
buibaoanh: As far as I know, `extra_part` is equivalent to "providerOptions" / "providerMetadata" in AI SDK. I added that alongside `extra_content` and `extra_body`.

apps/gateway/tests/utils/converters.test.ts
buibaoanh: Understand your concern, but I find pass-through logic is fragile:
- In javascript version, the options are in camelCase (I believe this is following typescript convention)
- In python version, it is snake_case

Pass-through snake_case options will fail validation in AI SDK, I am confused if gateway should support one version of providerOptions or both?

Below is what I understand from AI SDK:
image

apps/gateway/src/middlewares/models/gemini.ts
heiwen: 1/ LGTM

2/ Current design tightly couples a library-level function to application-level entities, that should be avoided => use built-in warnings.

buibaoanh:
CURL testing:
[hebo-gemini-test.md](https://github.com/user-attachments/files/24662988/hebo-gemini-test.md)

buibaoanh:
QA results: 
[hebo-gemini-test.md](https://github.com/user-attachments/files/24662782/hebo-gemini-test.md)

apps/gateway/src/middlewares/providers/vertex.ts
```diff
@@ -24,6 +33,78 @@ export class VertexProviderAdapter
...
+            ? this.transformProviderMetadata(part.providerMetadata)
+            : undefined,
+      }),
+    );
+    return {
+      ...result,
+      content: transformedContent,
+      providerMetadata: transformedProviderMetadata,
+    };
+  }
```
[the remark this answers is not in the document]
buibaoanh: result.content is part of the schema, it won't be null, check this https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-text#content

turisanapo (approved):

PR #141, opened 2026-01-05
Turisanapo/heb 243 auth include organizations backend services only

turisanapo:
## Summary by CodeRabbit

* **New Features**
  * Multi-organization support with teams, automatic organization creation on signup, and active-organization session handling
  * Organization invitation system with templated invitation emails and OTP improvements
  * Organization-scoped data for agents, branches, and provider configurations; team creation tied to agent creation

* **Other**
  * Per-request auth client and stricter auth validation requiring organization context
  * Consistent, shared slug generation utility and corresponding DB schema adjustments/migrations


apps/auth/src/better-auth.ts
```diff
@@ -25,6 +26,38 @@ function getCookieDomain() {
...
+  let membership = await prisma.members.findFirst({
+    where: { userId: newSession.user.id },
+  });
+
+  if (!membership) {
+    const org = await prisma.organizations.create({
+      data: {
+        id: crypto.randomUUID(),
+        name: `${newSession.user.name || newSession.user.email.split("@")[0]}'s Workspace`,
+        slug: `${newSession.user.id.slice(0, 8)}-workspace`,
```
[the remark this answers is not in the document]
turisanapo: Will handle in FUTURE

apps/api/prisma/schema.prisma
```diff
@@ -43,8 +42,8 @@ model agents {
   deleted_by String?
   /// @prismabox.hide
   deleted_at DateTime?
-
-  @@schema("api")
+  /// @prismabox.hide
+  organization_id String
```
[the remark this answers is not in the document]
turisanapo: This is as intended.

apps/api/prisma/schema.prisma
```diff
@@ -72,9 +71,10 @@ model branches {
   deleted_by String?
   /// @prismabox.hide
   deleted_at DateTime?
+  /// @prismabox.hide
+  organization_id String
```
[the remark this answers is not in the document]
turisanapo: This is as intended.

apps/api/prisma/schema.prisma
```diff
@@ -96,6 +96,6 @@ model provider_configs {
   deleted_by String?
   /// @prismabox.hide
   deleted_at DateTime?
-
-  @@schema("api")
+  /// @prismabox.hide
+  organization_id String
```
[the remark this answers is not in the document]
turisanapo: This is as intended.

apps/auth/prisma/schema.prisma
```diff
@@ -99,5 +96,49 @@ model apikeys {
...
+  teamId         String?       @db.Text
+  createdAt      DateTime      @default(now())
+  expiresAt      DateTime
+  inviter        users         @relation(fields: [inviterId], references: [id], onDelete: Cascade)
+  organization   organizations @relation(fields: [organizationId], references: [id], onDelete: Cascade)
+
+  @@index([email])
+  @@index([organizationId])
+  @@index([inviterId])
 }
```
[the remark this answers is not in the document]
turisanapo: The schema has been implemented as documented by better auth.

apps/auth/src/better-auth.ts
```diff
@@ -25,6 +27,59 @@ function getCookieDomain() {
...
+  const isNewUserPath =
+    ctx.path.startsWith("/callback/") || ctx.path === "/sign-in/email-otp";
+
+  const membership = isNewUserPath
+    ? await ensureUserHasOrganization(
+        newSession.user.id,
+        newSession.user.name,
+        newSession.user.email,
+      )
+    : await prisma.members.findFirst({ where: { userId: newSession.user.id } });
```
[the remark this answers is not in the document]
turisanapo: Will address later

apps/auth/src/better-auth.ts
```diff
@@ -25,26 +27,102 @@ function getCookieDomain() {
...
+  if (membership) {
+    await prisma.sessions.update({
+      where: { id: newSession.session.id },
+      data: { activeOrganizationId: membership.organizationId },
+    });
+  }
+
+  const teams = ctx.headers
+    ? await auth.api.listUserTeams({ headers: ctx.headers })
+    : [];
```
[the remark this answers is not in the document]
turisanapo: This is not happening.

apps/auth/src/better-auth.ts
```diff
@@ -83,6 +161,27 @@ export const auth = betterAuth({
...
+            sessions.map((session) =>
+              prisma.sessions.update({
+                where: { id: session.id },
+                data: { teamIds: [...(session.teamIds ?? []), team.id] },
+              }),
+            ),
+          );
+        },
+      },
+    }),
```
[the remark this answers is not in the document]
turisanapo: Rare event

apps/api/prisma/schema.prisma
```diff
@@ -61,20 +62,23 @@ model branches {
...
+  /// @prismabox.hide
+  deleted_by      String?
   /// @prismabox.hide
-  deleted_by String?
+  deleted_at      DateTime?
   /// @prismabox.hide
-  deleted_at DateTime?
+  organization_id String
+  /// @prismabox.hide
+  team_id         String
```
heiwen: Do we need the team_id on branches? Is this the same logic as organization_id, or does that logic not apply?

apps/api/src/modules/agents.ts
```diff
@@ -35,17 +38,35 @@ export const agentsModule = new Elysia({
   )
   .post(
     "/",
-    async ({ body, dbClient }) => {
+    async (ctx) => {
+      const { body, dbClient, request } = ctx;
+      const organizationId = (ctx as unknown as { organizationId: string })
```
heiwen: Nitpick: the unknown and string casts seem unnecessary, ctx should be properly typed in ElysiaJS. If it's necessary, there is probably a problem with the auth plugin.

apps/api/src/modules/agents.ts
```diff
@@ -35,17 +38,35 @@ export const agentsModule = new Elysia({
...
+        .organizationId;
+
+      const { data: team, error } = await authClient.organization.createTeam({
+        name: `${body.name}'s Team`,
+        organizationId,
+        fetchOptions: { headers: getAuthHeaders(request) },
```
heiwen: Can we set fetchOptions generally on the fetch object in the shared-api client, same as we do in the console?

apps/auth/src/better-auth.ts
```diff
@@ -25,26 +30,99 @@ function getCookieDomain() {
...
+
 export const auth = betterAuth({
   baseURL,
   basePath: "/v1",
   accountLinking: {
     enabled: true,
     trustedProviders: ["google", "github", "microsoft"],
   },
+  session: {
+    additionalFields: {
```
heiwen: Why do we need that?

apps/auth/src/better-auth.ts
```diff
@@ -25,26 +30,99 @@ function getCookieDomain() {
...
+}
+
+const afterHook = createAuthMiddleware(async (ctx) => {
+  const newSession = ctx.context.newSession;
+  if (!newSession) return;
+
+  const isNewUserPath =
+    ctx.path.startsWith("/callback/") || ctx.path === "/sign-in/email-otp";
+
+  const membership = isNewUserPath
```
heiwen: Open TODO: migrate to after user create hook.

apps/auth/src/better-auth.ts
```diff
@@ -25,26 +30,99 @@ function getCookieDomain() {
 }
 const cookieDomain = getCookieDomain();
 
+async function ensureUserHasOrganization(
```
heiwen: Suggest externalising larger functions with domain logic into lib/

packages/shared-api/utils/create-slug.ts
```diff
@@ -0,0 +1,26 @@
...
+  const suffix = cryptoRandomString({
+    length: suffixLength,
+    // eslint-disable-next-line no-secrets/no-secrets
+    characters: "abcdefghijklmnopqrstuvwxyz0123456789",
+  });
+
+  return base ? `${base}-${suffix}` : suffix;
+};
+
+export const createOrgSlug = (name: string | null, email: string): string => {
```
heiwen: 1/ `createOrgSlug` implies we can only use this for organizations. But it can be used for any name. `slugFormName` and `slugFromString` would probably be better names.
2/ We should probably add special character handling, particularly for `slugFromName`, but could be useful for edge cases in `slugFromString` as well.

apps/auth/src/lib/email.ts
```diff
@@ -31,84 +55,90 @@ export async function sendVerificationOtpEmail({
...
+  invitationId: string;
+  organizationName: string;
+  inviterName: string | null;
+  inviterEmail: string;
+  consoleUrl?: string;
+}) {
+  if (!isRemote) {
+    console.info(
+      ">>> Organization Invitation:",
+      `${consoleUrl || "http://localhost:5173"}/accept-invitation?id=${invitationId}`,
```
heiwen: This line should be able to reuse `new Url("/accept-invitation", consoleUrl)` from below instead of redefining the hostname:port.

packages/shared-api/package.json
```diff
@@ -33,6 +33,8 @@
     "@opentelemetry/exporter-trace-otlp-http": "^0.208.0",
     "@opentelemetry/sdk-trace-node": "^2.2.0",
     "better-auth": "^1.4.5",
+    "crypto-random-string": "^5.0.0",
```
heiwen, line 36: I see you are adding these two dependencies here, but not removing in api. If they are required in both places, then this one should be a peerDependency (+dev). If they are only required here, then they should be removed in API.

apps/api/src/modules/agents.ts
turisanapo: This is a know limitation of Elysia/Typescript we have found previously as well. The auth plugin is registered on the parent (index.ts) module, not here on the agents module. The type is not correctly inferred for this reason.

packages/shared-api/package.json
turisanapo: Done

apps/auth/src/lib/email.ts
turisanapo: Done

packages/shared-api/utils/create-slug.ts
turisanapo: Done

apps/auth/src/better-auth.ts
turisanapo: Done

apps/api/src/modules/agents.ts
turisanapo: Based on my understanding, Better Auth doesn't support a dynamic headers callback at the client level. The `fetchOptions` you set on `createAuthClient` are static. For server-side calls, I need to pass request-specific headers per call.

apps/auth/src/lib/email.ts
```diff
@@ -16,12 +16,36 @@ const transport = new SmtpTransport({
...
+        <tr>
+          <td style="padding-top:10px;">
+            <div style="border-top:1px solid #e2e8f0;margin:6px 0 10px;"></div>
+            <div style="font-size:12px;line-height:18px;color:#475569;text-align:center;">If you weren't expecting this email, you can safely ignore it.</div>
+          </td>
+        </tr>
+      </table>
+    </td>
+  </tr>
+</table>`;
```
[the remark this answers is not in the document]
turisanapo: They are not user provided

apps/api/src/modules/agents.ts
heiwen: If I remember correctly, the solution was to simply re-attach the plugin here as well. ElysiaJS will handle de-duplication automatically: https://elysiajs.com/key-concept.html#dependency

apps/api/src/middleware/db-client.ts
```diff
@@ -2,13 +2,17 @@ import { Elysia } from "elysia";
 
 import { createDbClient } from "~api/lib/db/client";
 
-// Note: Must be used after authService to ensure userId is set
+// Note: Must be used after authService to ensure userId and organizationId are set
 export const dbClient = new Elysia({
   name: "db-client",
 })
   .resolve(function resolveDbClient(ctx) {
+    const { organizationId, userId } = ctx as unknown as {
```
heiwen: Nitpick: the unknown and string casts seem unnecessary, ctx should be properly typed in ElysiaJS. If it's necessary, there is probably a problem with the auth plugin.

apps/api/src/modules/agents.ts
heiwen: That doesn't sound right. The auth client should be derived into the ElysiaJS instance, the same way that the dbClient is (instead of exported as a global singleton). When you initialize it, you can pass the fetchOptions to it once, instead of in every call.

apps/api/src/modules/agents.ts
```diff
@@ -35,13 +38,33 @@ export const agentsModule = new Elysia({
...
+      }
+
       return status(
         201,
         await dbClient.agents.create({
           data: {
             name: body.name,
-            slug: createSlug(body.name, true),
+            slug: agentSlug,
+            team_id: team.id,
```
heiwen, line 73: Let's keep team_id as implementation detail for now, until we really need it.

packages/shared-api/middlewares/auth/better-auth.ts
```diff
@@ -1,13 +1,28 @@
...
 
 import { BadRequestError } from "../../errors";
 
 const AUTH_URL = process.env.AUTH_URL || "http://localhost:3000";
 
-const authClient = createAuthClient({
+export const authClient = createAuthClient({
   baseURL: new URL("/v1", AUTH_URL).toString(),
+  plugins: [
+    organizationClient({
```
heiwen: Do we have to repeat the options in the client? If so, sounds like code that should be shared between server and client.

apps/api/src/modules/agents.ts
```diff
@@ -35,13 +38,33 @@ export const agentsModule = new Elysia({
   )
   .post(
     "/",
-    async ({ body, dbClient }) => {
+    async (ctx) => {
+      const { body, dbClient, request } = ctx;
```
heiwen: Nitpick: move the unpacking back into the method definition for easier readability.

apps/api/src/modules/agents.ts
turisanapo: Updated

apps/api/src/modules/agents.ts
turisanapo: Done

apps/auth/src/lib/email.ts
```diff
@@ -31,84 +55,89 @@ export async function sendVerificationOtpEmail({
...
-
-  if (!consoleUrl) {
-    console.warn("Missing origin header, cannot send verification email");
-    return;
+    if (!hasSmtpConfig()) return;
   }
+  if (!consoleUrl)
+    return console.warn(
+      "Missing origin header, cannot send verification email",
+    );
```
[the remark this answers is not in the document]
turisanapo: Warn it is fine and consistent with other log messages of this kind (expected behaviour in dev, but issue in prod)

apps/api/src/modules/agents.ts
```diff
@@ -35,13 +38,33 @@ export const agentsModule = new Elysia({
...
+        organizationId,
+        agentSlug,
+        fetchOptions: { headers: getAuthHeaders(request) },
+      });
+
+      if (error || !team) {
+        throw new Error(
+          `Failed to create team: ${error?.message ?? "Unknown error"}`,
+        );
+      }
```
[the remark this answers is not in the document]
turisanapo: Not relevant

packages/shared-api/utils/create-slug.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (!suffixLength) return base;
+
+  const suffix = cryptoRandomString({
+    length: suffixLength,
+    // eslint-disable-next-line no-secrets/no-secrets
+    characters: "abcdefghijklmnopqrstuvwxyz0123456789",
+  });
+
+  return base ? `${base}-${suffix}` : suffix;
+};
```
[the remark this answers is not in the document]
turisanapo: Nope, this is by design

apps/api/src/middleware/db-client.ts
turisanapo: Done

apps/api/src/modules/agents.ts
turisanapo: We will need a migration to fill the missing values if we don't start populating it from the start.

packages/shared-api/middlewares/auth/better-auth.ts
turisanapo: Done

apps/api/src/modules/agents.ts
turisanapo: Done

apps/auth/src/lib/email.ts
```diff
@@ -31,84 +55,92 @@ export async function sendVerificationOtpEmail({
...
+  email: string;
+  invitationId: string;
+  organizationName: string;
+  inviterName: string | null;
+  inviterEmail: string;
+  consoleUrl?: string;
+}) {
+  const acceptUrl = new URL(
+    "/accept-invitation",
+    consoleUrl || "http://localhost:5173",
```
heiwen: `consoleUrl` should automatically be localhost in local development environment, shouldn't it?

packages/shared-api/middlewares/auth/better-auth.ts
```diff
@@ -1,15 +1,46 @@
...
+      additionalFields: {
+        agentSlug: {
+          type: "string",
+        },
+      },
+    },
+  },
 });
 
+export const createAuthClient = (request: Request) => {
```
heiwen: Is there any need to export this?

Nitpick: for readability, since neither getAuthHeaders nor organizationClientPlugin is reused in other places, would just put it inline here.

packages/shared-api/middlewares/auth/better-auth.ts
```diff
@@ -25,21 +56,22 @@ export const authServiceBetterAuth = new Elysia({
...
-    });
+    const { data: session, error } = await authClient.getSession();
 
     if (error || !session) {
       log.info({ error }, "Authentication failed or no credentials provided");
-      return { userId: undefined } as const;
+      return {
+        organizationId: undefined,
+        userId: undefined,
+        authClient,
```
heiwen: Should authClient be undefined here as well?

apps/api/src/modules/agents.ts
heiwen: Misread the code. Understand now.

packages/shared-api/middlewares/auth/auth-service.ts
```diff
@@ -7,9 +7,10 @@ export const authService = new Elysia({ name: "auth-service" })
   .use(authServiceBetterAuth)
   .macro({
     isSignedIn: {
-      beforeHandle: function checkIsSignedIn({ userId }) {
-        if (!userId) throw new AuthError("Unauthorized");
+      beforeHandle: function checkIsSignedIn({ organizationId, userId }) {
+        if (!organizationId || !userId) throw new AuthError("Unauthorized");
       },
     },
+    // FUTURE: Add permission checks
```
heiwen: Permission checks won't be here, they will be very dependent on the actual business logic and live in the routes itself.

apps/auth/src/lib/email.ts
turisanapo: Right, updated.

packages/shared-api/middlewares/auth/better-auth.ts
turisanapo: I updated the imports.

packages/shared-api/middlewares/auth/better-auth.ts
turisanapo: No specific reason for it to be defined here. Updated for consistency.

packages/shared-api/middlewares/auth/auth-service.ts
turisanapo: Updated.

apps/auth/src/lib/organization.ts
```diff
@@ -0,0 +1,65 @@
...
+    { isolationLevel: "Serializable" },
+  );
+}
+
+export function createOrganizationHook(prisma: PrismaClient) {
+  return createAuthMiddleware(async (ctx) => {
+    const newSession = ctx.context.newSession;
+    if (!newSession) return;
+
+    const isNewUserPath =
```
heiwen: Wording is a bit confusing, suggest to rename to `isNewUser`

apps/auth/src/lib/organization.ts
```diff
@@ -0,0 +1,65 @@
...
+          newSession.user.id,
+          newSession.user.name,
+          newSession.user.email,
+        )
+      : // FUTURE: Define ordering of organizations
+        await prisma.members.findFirst({
+          where: { userId: newSession.user.id },
+        });
+
+    if (membership) {
```
heiwen: If membership is not set, that seems like an error condition that's being hidden right now?

apps/auth/src/lib/organization.ts
```diff
@@ -0,0 +1,65 @@
...
+      await prisma.sessions.update({
```
heiwen: 1/ Instead of doing an update, which is a second operation after the session create, can we use the session create hook as described here: https://www.better-auth.com/docs/plugins/organization#set-active-organization

2/ Additionally, can we "persist" the activeOrganization across sessions so that each individual gateway call doesn't have to query it again? https://github.com/better-auth/better-auth/issues/685

apps/auth/src/lib/organization.ts
turisanapo: Done

apps/auth/src/lib/organization.ts
turisanapo: Added explicit error

apps/auth/src/lib/organization.ts
turisanapo: Done

apps/auth/src/lib/organization.ts
```diff
@@ -0,0 +1,55 @@
...
+    await prisma.$transaction(async (tx) => {
+      const existing = await tx.members.findFirst({
+        where: { userId: newSession.user.id },
+      });
+      if (existing) return;
+
+      const org = await tx.organizations.create({
+        data: {
+          id: Bun.randomUUIDv7(),
+          name: `${newSession.user.name || newSession.user.email.split("@")[0]}'s Org`,
```
heiwen: Nitpick: why not the full-email?

apps/auth/src/lib/organization.ts
```diff
@@ -0,0 +1,55 @@
...
+      });
+      await tx.members.create({
+        data: {
+          id: Bun.randomUUIDv7(),
+          userId: newSession.user.id,
+          organizationId: org.id,
+          role: "owner",
+        },
+      });
+      await tx.sessions.update({
```
heiwen, line 37: Is this still needed with the createSessionHook?

apps/auth/src/better-auth.ts
```diff
@@ -83,6 +92,30 @@ export const auth = betterAuth({
...
+        },
+      },
+      async sendInvitationEmail(data, ctx) {
+        await sendOrganizationInvitationEmail({
+          email: data.email,
+          invitationId: data.id,
+          organizationName: data.organization.name,
+          inviterName: data.inviter.user.name,
+          inviterEmail: data.inviter.user.email,
+          consoleUrl: ctx?.headers.get("origin") ?? undefined,
```
heiwen, line 115: Nitpick: wouldn't `consoleUrl: ctx?.headers.get("origin")` be the same as `consoleUrl: ctx?.headers.get("origin") ?? undefined`

apps/auth/src/lib/organization.ts
turisanapo: Makes sense, updated

apps/auth/src/better-auth.ts
turisanapo: `?? undefined` is needed. `headers.get()` returns `string | null` (`null` when missing), while the function expects `string | undefined`. The `?? undefined` converts `null` to `undefined`.

apps/auth/src/lib/organization.ts
turisanapo: Yes, it is still needed to update the existing session after the organization is created at the first user signup. Without it, the session won't have a value for `activeOrganizationId` after the organization is created making the experience break since the app expects `activeOrganizationId` to have a valid value.

PR #149, opened 2026-01-15
Turisanapo/heb 281 infra fix preview deployments

turisanapo:
## Summary by CodeRabbit

* **Chores**
  * Standardized stage naming across API, auth, console, DB, gateway, and MCP for consistent domain construction.
  * Added normalized stage identifiers (lowercase, dash-separated) for reliability.
  * Added an ESLint rule exception for a specific infra configuration file.

* **Bug Fixes**
  * Improved secret retrieval to preserve true undefined values instead of returning the literal string "undefined".

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -129,6 +129,12 @@ const eslintConfig = [
       'unicorn/no-abusive-eslint-disable': 'warn'
     },
   },
+  {
+    files: ["infra/stacks/env.ts"],
+    rules: {
+      "sonarjs/slow-regex": "off",
+    },
+  },
 ];
 
 export default eslintConfig;
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -129,12 +129,6 @@ const eslintConfig = [
       'unicorn/no-abusive-eslint-disable': 'warn'
     },
   },
-  {
-    files: ["infra/stacks/env.ts"],
-    rules: {
-      "sonarjs/slow-regex": "off",
-    },
-  },
 ];
 
 export default eslintConfig;
--- a/infra/stacks/api.ts
+++ b/infra/stacks/api.ts
@@ -1,9 +1,9 @@
 import heboAuth from "./auth";
 import heboCluster from "./cluster";
 import heboDatabase, { createMigrator } from "./db";
-import { otelSecrets, isProd, formattedStage } from "./env";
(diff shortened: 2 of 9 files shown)


eslint.config.mjs
```diff
@@ -129,6 +129,12 @@ const eslintConfig = [
       'unicorn/no-abusive-eslint-disable': 'warn'
     },
   },
+  {
```
heiwen: Specifying file-specific excludes in the global eslint.config.mjs is discouraged. Why are you not putting an `// eslint-disable` in the file directly?

infra/stacks/env.ts
```diff
@@ -29,7 +30,10 @@ export const vertexProject = getSstSecret("VertexProject");
 export const vertexLocation = getSstSecret("VertexLocation");
 
 // OTEL Exporter
-export const grafanaEndpoint = getSstSecret("GrafanaEndpoint");
+export const grafanaEndpoint = getSstSecret(
+  "GrafanaEndpoint",
+  "http://localhost:3000",
```
heiwen: 1/ The specific host:port is confusing, since it overlaps with our auth service in development?
2/ Can we fix this at the point where the URL is consumed instead, so that it parses the string `"undefined"` to the value `undefined` (e.g. in secrets.ts)?

infra/stacks/api.ts
```diff
@@ -1,9 +1,9 @@
 import heboAuth from "./auth";
 import heboCluster from "./cluster";
 import heboDatabase, { createMigrator } from "./db";
-import { otelSecrets, isProd } from "./env";
+import { otelSecrets, isProd, formattedStage } from "./env";
```
heiwen: Nitpick: "formatting" is for console / UI output. In this case, it's "normalized"

infra/stacks/env.ts
```diff
@@ -66,3 +70,9 @@ export const otelSecrets = [
 ];
 
 export const isProd = $app.stage === "production";
+export const formattedStage = $app.stage
+  .trim()
+  .toLowerCase()
+  .replaceAll(/[^a-z0-9]+/g, "-")
+  .replaceAll(/^-+/g, "")
```
heiwen: These two can be combined into a single replace: .replaceAll(/(^-|-$)/g, "");

PR #150, opened 2026-01-17
chores(infra): remove IS_REMOTE

turisanapo:
## Summary by CodeRabbit

* **New Features**
  * Expose a configurable auth URL and add root-domain detection for cross-service use.

* **Refactor**
  * CORS and cookie domain handling now derive root domains for consistent multi-domain behavior.
  * Production detection standardized to NODE_ENV; package export surface updated to expose env utilities.

* **Bug Fixes**
  * Email logging and SMTP validation now respect production mode.
  * Container metadata credentials are fetched when available.

* **Chores**
  * Environment variables adjusted across deployment configs for consistent NODE_ENV handling.


packages/shared-api/utils/cors.ts
```diff
@@ -1,15 +1,26 @@
...
-      }
-    : {
-        origin: true,
-        credentials: true,
-        maxAge: 3600,
-      };
+import { authUrl } from "../env";
+import { getRootDomain } from "./root-domain";
+
+const rootDomain = getRootDomain(authUrl);
```
heiwen: Nitpick: Do the escaping here:

`const rootDomain = getRootDomain(authUrl)?.replaceAll(".", "\\.")`

packages/shared-api/utils/cors.ts
```diff
@@ -1,15 +1,26 @@
...
+
+export const corsConfig = rootDomain
+  ? {
+      // Matches HTTPS origins for exact domain or subdomains
+      // eslint-disable-next-line security/detect-non-literal-regexp
+      origin: new RegExp(
```
heiwen, line 9: Nitpick: with escaping done before, you can simplify this to:

`origin: new RegExp(^https://(?:${rootDomain}|(?:[a-z0-9-]+\.)+${rootDomain})$, "i")`

(`String.raw` should not be required)

packages/shared-api/utils/domains.ts
heiwen, line 1: Reminder: don't name a module the same as the single function it contains. Instead, expect the module to grow over time, with functions of related purposes. In this case, this would be something like `domains.ts`, to indicate utilities to handle domains.

apps/auth/src/better-auth.ts
```diff
@@ -11,33 +13,25 @@ import {
...
-  baseURL,
+  baseURL: authUrl,
   basePath: "/v1",
   accountLinking: {
     enabled: true,
     trustedProviders: ["google", "github", "microsoft"],
   },
   advanced: {
-    useSecureCookies: isRemote,
+    useSecureCookies: isProd,
```
heiwen: 1/ Is this correct? Auth in preview environments is running on TLS as well, isn't it? Would it still work without useSecureCookies?

2/ If it is correct, to my understanding of the documentation betterAuth automatically sets this behavior, meaning we can remove that explicit configuration option: https://www.better-auth.com/docs/concepts/cookies#secure-cookies

packages/shared-api/utils/cors.ts
turisanapo: Updated. However, the formatter is automatically replacing `"\\."` with `String.raw`.

packages/shared-api/utils/cors.ts
turisanapo: Updated

packages/shared-api/utils/domains.ts
turisanapo: Updated

apps/auth/src/better-auth.ts
```diff
@@ -11,33 +13,25 @@ import {
...
-// Set to the eTLD+1 (e.g., "hebo.ai") so auth cookies flow to api/gateway.
-function getCookieDomain() {
-  const { hostname } = new URL(baseURL);
-  return hostname === "localhost"
-    ? undefined
-    : hostname.split(".").slice(-2).join(".");
-}
-const cookieDomain = getCookieDomain();
+// Set to the eTLD+1 (e.g., "example.com") so auth cookies flow to api/gateway.
+const cookieDomain = getRootDomain(authUrl);
```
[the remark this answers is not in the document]
turisanapo: Not an expected issue currently.

apps/auth/src/better-auth.ts
turisanapo: You are right, we would replicate the same behaviour as the default one. Updated.

PR #161, opened 2026-01-27
feat(auth): enable cookie cache for clients

turisanapo:
## Summary by CodeRabbit

* **New Features**
  * Session cookie caching enabled for faster auth checks.
  * Sign-in redirect now routes through "/?after-signin" for clearer post-login flow.

* **Improvements**
  * Robust cookie-based session validation with proactive sign-out on failure.
  * User initialization streamlined to avoid redundant session fetches; user initials derived automatically.

* **Bug Fixes**
  * Removed dynamic cross-subdomain cookie logic to simplify cookie behavior.

* **Chores**
  * Consolidated and updated import/module paths across services.


apps/auth/src/better-auth.ts
```diff
@@ -32,6 +32,10 @@ export const auth = betterAuth({
       enabled: Boolean(cookieDomain),
       domain: cookieDomain,
     },
+    // FUTURE: restore default once we have a better way to check for session cookie (SSR?)
+    defaultCookieAttributes: {
+      httpOnly: false,
+    },
```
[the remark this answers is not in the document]
turisanapo: Updated FUTURE.

apps/console/app/lib/auth/better-auth.ts
```diff
@@ -38,16 +38,33 @@ const authClient = createAuthClient({
...
+    const isComingFromSignIn = document.referrer.includes("/signin");
+
+    const session = await authClient.getSession({
+      query: { disableCookieCache: isComingFromSignIn },
+    });
+    const user = session.data?.user as User | undefined;
+    if (!user) {
+      redirectToSignIn();
+      return;
+    }
```
heiwen: When would this path trigger? Feels like an error condition / potential endless loop?

apps/console/app/lib/auth/better-auth.ts
```diff
@@ -38,16 +38,33 @@ const authClient = createAuthClient({
   ],
 });
 
+const redirectToSignIn = () => {
```
heiwen: The method does more than the name implies. Also, on the second call site, `shellStore.user` is already undefined. Suggest to re-inline the code.

apps/console/app/lib/auth/better-auth.ts
turisanapo: No, it will not introduce a loop since after the user succesfully signs-in, they will create a valid session.

This case is to make sure we are handling users for whom we manually revoked the session from the DB (for security reasons or stop some exploitation of our resources, for example)

heiwen: Understand the scenario now, but I don't believe it's handled. The code won't be triggered, as long as the cookie still exists and shellStore is set. In general, after session revocation, any call to any service (api, gateway, ...) would silently fail with an auth error. So that's technically where it would have to be handled, if we want to cover that edge case.

packages/shared-api/middlewares/auth/better-auth.ts
```diff
@@ -52,6 +56,19 @@ export const authServiceBetterAuth = new Elysia({
...
+      const cachedSession = await getCookieCache(ctx.request, {
+        secret: authSecret,
+      });
+      if (cachedSession) {
+        return {
+          organizationId: cachedSession.session.activeOrganizationId,
+          userId: cachedSession.user.id,
+          authClient,
+        } as const;
+      }
```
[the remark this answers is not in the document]
turisanapo: The fallback is specifically for api keys and does not happen for cookie.

apps/console/app/lib/auth/better-auth.ts
turisanapo: Done.

apps/console/app/lib/auth/better-auth.ts
turisanapo: Updated as discussed.

apps/console/app/lib/auth/better-auth.ts
```diff
@@ -38,16 +38,33 @@ const authClient = createAuthClient({
...
+    if (!document.cookie.includes("better-auth.session_token=")) {
+      redirectToSignIn();
+      return;
+    }
 
-    if (!user) {
-      globalThis.location.replace("/signin");
+    if (shellStore.user) {
       return;
     }
```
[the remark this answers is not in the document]
turisanapo: This is not valid. It is a trade-off we accept

apps/console/app/lib/auth/better-auth.ts
```diff
@@ -40,14 +41,30 @@ const authClient = createAuthClient({
 
 export const authService: AuthService = {
   async ensureSignedIn() {
-    const session = await authClient.getSession();
-    const user = session.data?.user as User | undefined;
-
-    if (!user) {
+    const headers = new Headers();
+    headers.set("cookie", document.cookie);
+    if (!getSessionCookie(headers)) {
```
heiwen, line 45: Nitpick, initialise Headers in the constructor:

```
const headers = new Headers({ cookie: document.cookie }))
`if (!getSessionCookie(headers)) {`
```

apps/console/app/lib/auth/better-auth.ts
```diff
@@ -40,14 +41,30 @@ const authClient = createAuthClient({
...
+      return;
+    }
+
+    // Disable cookie cache only after fresh sign-in to ensure we get the latest session
+    const url = new URL(globalThis.location.href);
+    const isComingFromSignIn = url.searchParams.has("after-signin");
+    if (isComingFromSignIn) {
+      url.searchParams.delete("after-signin");
+      globalThis.history.replaceState({}, "", url.pathname + url.search);
+    }
```
heiwen: Why exactly do we need to delete the search param and rewrite the history?

packages/shared-api/middlewares/auth.ts
```diff
@@ -52,10 +59,30 @@ export const authServiceBetterAuth = new Elysia({
...
+      // Clear the session cookie when unauthorized
+      const { sameSite, ...restAttributes } =
+        cookieConfig.sessionToken.attributes;
+      ctx.set.cookie ??= {};
+      ctx.set.cookie[cookieConfig.sessionToken.name] = {
+        value: "",
+        ...restAttributes,
+        sameSite: sameSite.toLowerCase() as "lax" | "strict" | "none",
+        maxAge: 0,
+      };
```
heiwen: Nitpick: simplified version:

```
const { attributes, name } = cookieConfig.sessionToken;

(ctx.set.cookie ??= {})[name] = {
  value: "",
  ...attributes,
  maxAge: 0,
};
```

I believe there shouldn't be a need to lowercase `sameSite`, as it should be correctly returned by `getCookies(...)`

packages/shared-api/middlewares/auth/cookie-options.ts
```diff
@@ -0,0 +1,24 @@
+import { authUrl } from "../../env";
+import { getRootDomain } from "../../utils/domains";
+
+// eTLD+1 domain for cross-subdomain cookies (e.g., "hebo.ai")
+const cookieDomain = getRootDomain(authUrl);
+
+// Use secure cookies when not on localhost (production)
+const useSecureCookies = cookieDomain !== undefined;
```
heiwen: 1/ We previously agreed to consistency use process.env.NODE_ENV === "production" here
2/ For readability, we generally agreed to not create const for variables that are only read once, but inline them instead

packages/shared-api/lib/cookie-options.ts
```diff
@@ -0,0 +1,24 @@
+import { authUrl } from "../../env";
+import { getRootDomain } from "../../utils/domains";
+
+// eTLD+1 domain for cross-subdomain cookies (e.g., "hebo.ai")
+const cookieDomain = getRootDomain(authUrl);
+
+// Use secure cookies when not on localhost (production)
+const useSecureCookies = cookieDomain !== undefined;
+
+export const betterAuthCookieOptions = {
```
heiwen, line 7: Do we really need to explicitly add options which have default values, or can we rely on the previous behavior that defaults are handled by betterAuth internally? I guess in this case the question would be whether the `getCookies()` function properly backfill the internal defaults?

https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/cookies/index.ts#L34

packages/shared-api/lib/cookie-options.ts
heiwen, line 1: Structure: The `auth` service should not be consuming the cookie options from the `middleware` (since it doesn't use the middleware), instead `lib` would be the right folder. The `middleware` then also consumes it from `lib`. Similar to how DB connection strings also live in `lib`.

Unrelated to this PR:
1/ `cors` and `otel` should be in `lib` as well, since they define actual configuration / behaviour
2/ Now that we don't have a localhost service anymore, `better-auth` should be merged into `auth-service` and since that only leaves one module in the `auth` folder it can be moved up one level to `auth.ts`

apps/console/app/lib/auth/better-auth.ts
turisanapo: Good one, we don't need to. Simplified.

apps/console/app/lib/auth/better-auth.ts
turisanapo: Done

packages/shared-api/middlewares/auth.ts
turisanapo: Unfortunately, the `toLowerCase()` call is still required due to a type mismatch between Better Auth and Elysia:

- Better Auth's getCookies returns `sameSite` with type "Strict" | "Lax" | "None" | "strict" | "lax" | "none"
- Elysia's cookie type only accepts "lax" | "strict" | "none" (lowercase only)

packages/shared-api/middlewares/auth/cookie-options.ts
turisanapo: I removed `useSecureCookies` since it is possible to use the default value.

packages/shared-api/lib/cookie-options.ts
turisanapo: Good catch, removed.

packages/shared-api/lib/cookie-options.ts
turisanapo: Updated.

apps/console/app/lib/auth/better-auth.ts
```diff
@@ -40,14 +41,25 @@ const authClient = createAuthClient({
...
+
+    // Disable cookie cache only after fresh sign-in to ensure we get the latest session
+    const isComingFromSignIn = new URL(
+      globalThis.location.href,
+    ).searchParams.has("after-signin");
+    const session = await authClient.getSession({
+      query: { disableCookieCache: isComingFromSignIn },
+    });
+    const user = session.data?.user as User;
```
heiwen, line 62: Is this cast really necessary? (It wasn't before?)

[In general, casts are orange flags for potential runtime errors that happen later down the road and should be avoided if possible.)

packages/shared-api/middlewares/auth.ts
```diff
@@ -80,3 +106,14 @@ export const authServiceBetterAuth = new Elysia({
     } as const;
   })
   .as("scoped");
+
+export const authService = new Elysia({ name: "auth-service" })
```
heiwen: Do we need two nested Elysia instance here, or can the authServiceBetterAuth simply be merged into this one?

packages/shared-api/middlewares/auth.ts
heiwen: In that case you can probably cast the whole object to the ElysiaJS version, instead of having to merge-in and cast a single property. Also, if I see it correctly, `cookie` is directly available on the `ctx`, meaning you don't have to go through `ctx.set`.

```
const { attributes, name } = cookieConfig.sessionToken;

ctx.cookie[name] = {
  value: "",
  maxAge: 0,
  ...attributes,
} as Cookie<string>;
```

No `toLowerCase()` needed, since we know that the actual values in the cookies from better-auth are valid. It just means that either (i) their type is too lax or (ii) Elysia's type is too strict. Either-way, it's just a type thing, not a runtime issue.

turisanapo: Good one! Updated

apps/console/app/lib/auth/better-auth.ts
turisanapo: There's a mismatch between the user type returned by Better Auth and the User type we use in `shellStore`:

- `image` can be null or undefined in Better Auth, while `shellStore` only accepts undefined when missing
- `initials` is only present in `shellStore`. It is a computed field to avoid recalculating initials where needed

I removed the cast by being explicit in the conversion, which avoids hiding potential runtime errors. An alternative would be to use Better Auth's user type directly in the store and compute initials where required

packages/shared-api/middlewares/auth.ts
turisanapo: Updated

PR #168, opened 2026-02-03
chores(gateway): simplify gateway logics by integrating hebo ai gateway

turisanapo:
## Summary by CodeRabbit

* **New Features**
  * Public gateway endpoints for models, chat completions, and embeddings; centralized model/provider resolution with caching and dynamic provider creation; gateway configuration export.

* **Bug Fixes**
  * Embeddings badge display corrected.
  * Secret retrieval behavior standardized across services.
  * AWS credential injection flow fixed.
  * DB query relaxed to return null instead of throwing when provider config is missing.

* **Chores**
  * Dependency upgrades; added API/GATEWAY env vars and dynamic OpenAPI config; README updated.

* **Removed**
  * Legacy model/provider adapters, conversion utilities, OpenAI-compatible schemas, related routes and tests.


packages/shared-api/utils/secrets.ts
```diff
@@ -1,16 +1,12 @@
 import { secrets } from "bun";
 import { Resource } from "sst";
 
-export const getSecret = async (name: string, required = true) => {
+export const getSecret = async (name: string) => {
```
heiwen, line 4: Can you explain the rational behind this change? Does it still allow for deploying without having to specify all secrets (like it was the intent before for preview environments)?

apps/gateway/package.json
```diff
@@ -15,23 +15,23 @@
...
   "dependencies": {
-    "@ai-sdk/amazon-bedrock": "^3.0.73",
-    "@ai-sdk/google-vertex": "^3.0.97",
-    "@ai-sdk/groq": "^2.0.34",
-    "@ai-sdk/provider": "2.0.1",
+    "@ai-sdk/amazon-bedrock": "^4.0.46",
+    "@ai-sdk/google-vertex": "^4.0.41",
+    "@ai-sdk/groq": "^3.0.21",
+    "@ai-sdk/provider": "3.0.7",
     "@ai-sdk/provider-utils": "catalog:",
```
heiwen: Cleanup: `@ai-sdk/provider-utils` shouldn't be needed anymore

apps/gateway/package.json
```diff
@@ -15,23 +15,23 @@
...
     "@aws-sdk/client-bedrock": "^3.968.0",
     "@aws-sdk/credential-providers": "^3.980.0",
     "@bogeychan/elysia-logger": "catalog:",
     "@elysiajs/cors": "catalog:",
-    "@elysiajs/openapi": "catalog:",
     "@elysiajs/opentelemetry": "catalog:",
+    "@hebo-ai/gateway": "^0.2.0",
     "@hebo/api": "workspace:*",
     "@hebo/shared-api": "workspace:*",
     "ai": "catalog:",
```
heiwen: Cleanup: `ai` shouldn't be needed anymore

apps/gateway/package.json
```diff
@@ -15,23 +15,23 @@
...
     "elysia": "catalog:",
     "quick-lru": "^7.3.0",
-    "voyage-ai-provider": "^2.0.0",
+    "voyage-ai-provider": "^3.0.0",
     "zod": "catalog:"
```
heiwen: Cleanup: `zod` shouldn't be needed anymore

apps/gateway/src/index.ts
```diff
@@ -23,28 +22,19 @@ export const createGateway = () =>
...
+        .use(dbClient)
+        .post(
+          `/*`,
+          ({ request, dbClient }) => gw.handler(request, { dbClient }),
+          { parse: "none" },
+        ),
     )
     // Public routes (no authentication required)
-    .group("/v1", (app) => app.use(models));
+    .mount("/v1/models", gw.routes["/models"].handler);
```
heiwen: You're now mounting this twice, once under `"/*"` and once under `/v1/models`. I would expect the "*" route to catch it first (since it's first and star). Can you confirm that it is really available without auth like this?

apps/gateway/src/index.ts
```diff
@@ -23,28 +22,19 @@ export const createGateway = () =>
...
-      "/v1",
-      {
-        isSignedIn: true,
-      },
-      (app) => app.use(completions).use(embeddings),
+    .use(authService)
+    .group(basePath, { isSignedIn: true }, (app) =>
+      app
+        .use(dbClient)
+        .post(
```
heiwen, line 83: Using `post` is brittle. In the future, we might introduce `get` methods as well. Use `all` instead.

apps/gateway/src/gateway-config.ts
```diff
@@ -0,0 +1,74 @@
...
+const secrets = await loadProviderSecrets();
+
+export { basePath };
+
+const withPricing = (freeTokens: number) => ({
+  additionalProperties: { pricing: { monthly_free_tokens: freeTokens } },
+});
+
+const getResolver = (ctx: { state: unknown }) =>
+  new ModelResolver(
```
heiwen: If you need to create a new `ModelResolver(...)` every time `getResolver(...)` is called, it indicates that this is an unnecessary abstraction level.

apps/gateway/src/gateway-config.ts
```diff
@@ -0,0 +1,74 @@
...
+    (ctx.state as { dbClient: ReturnType<typeof createDbClient> }).dbClient,
+  );
+
+export const gw = gateway({
+  basePath,
+
+  providers: {
+    groq: createProvider("groq", { apiKey: secrets.groqApiKey }),
```
heiwen, line 26: Vercel AI SDK providers only check whether credentials are valid (or even filled with values), once you actually call a model. In order to keep our previous behaviour, whereas providers are not initialised in dev environments when secrets are not provided, you need to conditionally configure the providers here.

apps/gateway/src/gateway-config.ts
```diff
@@ -0,0 +1,74 @@
...
+  models: defineModelCatalog([
+    gptOss20b({
+      providers: ["bedrock", "groq"],
+      ...withPricing(12_000_000_000),
+    }),
+    gptOss120b({
+      providers: ["bedrock", "groq"],
+      ...withPricing(6_000_000_000),
+    }),
+    ...gemini["v3.x"].map((preset) =>
```
heiwen: "..." should not be required.

apps/gateway/src/gateway-config.ts
```diff
@@ -0,0 +1,74 @@
...
+  ]),
+
+  hooks: {
+    resolveModelId: async (ctx) =>
+      ctx.modelId && getResolver(ctx).resolveModelId(ctx.modelId),
+
+    resolveProvider: async (ctx) =>
+      ctx.modelId &&
+      getResolver(ctx).resolveProvider(
+        ctx.resolvedModelId ?? ctx.modelId,
```
heiwen: You can always use `resolvedModelId`.

apps/gateway/src/services/model-resolver.ts
```diff
@@ -0,0 +1,185 @@
...
+      const { apiKey } = config as ApiKeyProviderConfig;
+      return withCanonicalIdsForVoyage(createVoyage({ apiKey }));
+    }
+    default: {
+      throw new Error(`Unsupported provider: ${slug}`);
+    }
+  }
+}
+
+export class ModelResolver {
```
heiwen: In the current approach, both `resolveModelId` and `resolveProvider` lead to database calls. Instead you can do the call once in `resolveModelId` and pass the information via the `ctx.state` object from hook to hook.

apps/gateway/src/services/model-resolver.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  async resolveProvider(
+    modelId: string,
+    aliasPath: string,
+    defaultProviders: ProviderRegistry,
+  ): Promise<ProviderV3 | undefined> {
+    // Refresh AWS credentials for Vertex WIF before using Google models
+    if (modelId.startsWith("google/")) {
+      await injectMetadataCredentials();
```
heiwen: I guess this is a reasons why calls through Hebo gateway are still slow. Can this happen once on gateway start instead of in every call?

apps/gateway/src/services/model-resolver.ts
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    const config = await this.getModelConfig(aliasPath);
+
+    if (config.customProviderSlug) {
+      return this.getCustomProvider(config.customProviderSlug, modelId);
+    }
+
+    // Fallback to groq for openai/* models when bedrock not configured
```
heiwen: This can / should be handled by conditionally initialising providers on the config, depending on availability of secrets.

apps/gateway/src/services/model-resolver.ts
```diff
@@ -0,0 +1,185 @@
...
+    bedrockRegion,
+    voyageApiKey,
+    vertexServiceAccountEmail,
+    vertexAudience,
+    vertexLocation,
+    vertexProject,
+  };
+}
+
+export function createProvider(
```
heiwen: Given this is used by the gateway config AND model-resolver, it should conceptually live in a separate module.

apps/gateway/src/gateway-config.ts
```diff
@@ -0,0 +1,74 @@
...
+      ...withPricing(6_000_000_000),
+    }),
+    ...gemini["v3.x"].map((preset) =>
+      preset({ providers: ["vertex"], ...withPricing(0) }),
+    ),
+    voyage35({ providers: ["voyage"], ...withPricing(0) }),
+  ]),
+
+  hooks: {
+    resolveModelId: async (ctx) =>
```
heiwen: Unnecessary closure, just match the function signatures between the hook and your custom code.

apps/gateway/src/gateway-config.ts
```diff
@@ -0,0 +1,74 @@
...
+      ctx.modelId && getResolver(ctx).resolveModelId(ctx.modelId),
+
+    resolveProvider: async (ctx) =>
```
heiwen: Unnecessary closure, just match the function signatures, between the hook and your custom code.

apps/gateway/src/gateway-config.ts
```diff
@@ -0,0 +1,74 @@
...
+  },
+
+  models: defineModelCatalog([
+    gptOss20b({
+      providers: ["bedrock", "groq"],
+      ...withPricing(12_000_000_000),
+    }),
+    gptOss120b({
+      providers: ["bedrock", "groq"],
+      ...withPricing(6_000_000_000),
```
heiwen: `withPricing` reads confusing, looks like gptOSS actually costs money, but gemini is for free. But it's the opposite.

packages/shared-api/utils/secrets.ts
turisanapo: Yes, the behaviour is the same. Now, all the secrets are not required, so I just removed the option and made the wanted behaviour the default one.

apps/gateway/src/index.ts
turisanapo: Yes, I'm able to `GET /v1/models` without auth headers.

heiwen: I guess that's because the other routes are only mounted as "post", see my other comment how this is brittle.

packages/shared-api/utils/secrets.ts
heiwen: Clear

apps/gateway/src/index.ts
turisanapo: Tested again with `.all()`, the behaviour is consistent.

apps/gateway/src/services/model-resolver.ts
turisanapo: Based on the experiments I ran to investigate latency, this exact step takes between 1 and 4 ms. It is an internal HTTP call to the ECS task container.

We can't do it only once because ECS credentials are temporary and expire after 6 hours.

We could do some optimisation now, but given the operation being very cheap, I preferred avoid complexity.

apps/gateway/package.json
turisanapo: Good catch. Updated

apps/gateway/package.json
turisanapo: Good catch. Updated

apps/gateway/package.json
turisanapo: Good catch. Updated

apps/gateway/src/index.ts
turisanapo: Good point, updated

apps/gateway/src/gateway-config.ts
turisanapo: I included it to simplify the typing. However, in this last version, I'm matching signatures so no additional typing is required.

apps/gateway/src/gateway-config.ts
turisanapo: Updated

apps/gateway/src/gateway-config.ts
turisanapo: Updated

apps/gateway/src/services/model-resolver.ts
turisanapo: Updated

apps/gateway/src/index.ts
heiwen: Are you sure it captures this route and not the one previously registered by the .all() handler?

apps/gateway/src/services/model-resolver.ts
```diff
@@ -0,0 +1,115 @@
...
+
+import { injectMetadataCredentials } from "./aws-wif";
+import { createProvider } from "./provider-factory";
+
+import type { ProviderV3 } from "@ai-sdk/provider";
+import type { HookContext } from "@hebo-ai/gateway";
+
+const providerCache = new QuickLRU<string, ProviderV3>({ maxSize: 100 });
+
+export async function loadProviderSecrets() {
```
heiwen: This should be in provider-factory, shouldn't it?

apps/gateway/src/services/model-resolver.ts
```diff
@@ -0,0 +1,115 @@
...
+  if (modelId.startsWith("google/")) {
+    await injectMetadataCredentials();
+  }
+
+  const { customProviderSlug } = state.modelConfig as {
+    customProviderSlug?: ProviderSlug;
+  };
+
+  if (customProviderSlug) {
+    const { value: config } =
```
heiwen: Can you query the unredacted version in resolveModelId already, in order to avoid two db calls?

apps/gateway/src/gateway-config.ts
```diff
@@ -0,0 +1,59 @@
...
+  loadProviderSecrets,
+  resolveModelId,
+  resolveProvider,
+} from "./services/model-resolver";
+import { createProvider } from "./services/provider-factory";
+
+const basePath = "/v1";
+const secrets = await loadProviderSecrets();
+
+export { basePath };
```
heiwen: Nitpick: just define the export when defining the const.

apps/auth/src/better-auth.ts
```diff
@@ -99,23 +99,23 @@ export const auth = betterAuth({
       },
     }),
   ],
-  secret: await getSecret("AuthSecret", false),
+  secret: await getSecret("AuthSecret"),
```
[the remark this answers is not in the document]
turisanapo: The service startup will fail anyway and rolled back if this will ever happen

apps/gateway/src/services/model-resolver.ts
turisanapo: We would still need to execute two queries since the data resides in two different tables with no explicit relation.

We can reduce it to a single round-trip to the database by executing a `JOIN` query, but this would require writing raw SQL.

An alternative would be to eagerly fetch all the custom providers of a given organisation and then pick the matching one on the application layer. This would allow us to execute the two queries in parallel.

I would go with the second approach, since the number of total custom providers for an organisation is very limited.

apps/gateway/src/services/model-resolver.ts
turisanapo: Yes, moved.

apps/gateway/src/gateway-config.ts
turisanapo: Done, thanks

apps/gateway/src/gateway-config.ts
turisanapo: Updated

apps/gateway/src/index.ts
turisanapo: If I remove 

```
.mount("/v1/models", gw.routes["/models"].handler);
```

I get 

```
{
    "error": {
        "message": "Unauthorized",
        "type": "invalid_request_error",
        "code": "auth_error"
    }
}
```

apps/gateway/src/services/model-resolver.ts
heiwen: Understand, that's fine then. But I guess, we can move this code after the cache check?

apps/gateway/src/gateway-config.ts
heiwen: I can't find the conditional creation of the provider based on whether secrets are set or not. Where can I find it?

turisanapo: On `provider-factory`, `createProvider` now returns `undefined` when no secrets are passed.

apps/gateway/src/services/model-resolver.ts
turisanapo: The cache key depends on the result of the query; this is to miss the cache in case the DB configuration changes by design.

Alternatively, during configuration, a user who makes a mistake in setting up keys will update the configuration, but the change won't be acknowledged by the service, which would have been caching the older configuration.

Also, our cache is based on simple LRU, meaning that we can't even give a time expectation to the user if we still decide to proceed this way.

apps/gateway/src/index.ts
heiwen: Ok, it's working. But it is still confusing. Usually earlier routes are executed first. Suggest to change the order for clarity. And instead of `mount`, use `all` here as well, since it is a single route. I guess once you attach the schemas, you'll have to do that anyway.

apps/gateway/src/gateway-config.ts
heiwen: 1/ Is it though, it checks whether the config object is undefined, not the properties in the config object
2/ I'm not sure hebo-gateway can handle `undefined`, have you tested that?

apps/gateway/src/services/model-resolver.ts
heiwen: I guess this is "optimized" for the path where the credentials change frequently. But usually they are the same. How to optimize for the more common path?

Independently, for security reasons, suggest to hash the config key, instead of plain text credentials in there

apps/gateway/src/index.ts
turisanapo: Updated

apps/gateway/src/gateway-config.ts
turisanapo: 1/ good catch, updated
2/ I tested it, and the service does not break, however models related to undefined providers are still listed.

apps/gateway/src/services/model-resolver.ts
turisanapo: You're right that credentials rarely change. To skip the DB query on cache hit, we'd need either TTL-based caching (accepting short staleness) or event-driven invalidation when configs update.

I would leave these improvements for when we scale more, so that we will be able to pick the correct bottlenecks to improve performance.

turisanapo: Regarding the hash, updated.

apps/gateway/src/services/provider-factory.ts
```diff
@@ -0,0 +1,105 @@
...
+    vertexProject,
+  };
+}
+
+export function createProvider(
+  slug: ProviderSlug,
+  config: unknown,
+): ProviderV3 | undefined {
+  console.log("config", config);
+  switch (slug) {
```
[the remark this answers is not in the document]
turisanapo: fixed

apps/gateway/src/services/model-resolver.ts
heiwen: Since this affects the happy path of 3cat directly, would like to address now. Let's do TTL caching with short staleness.

apps/gateway/src/gateway-config.ts
heiwen: 2/ If models related to undefined providers are still listed, than calls will be routed there. If the providers doesn't have credentials, it will fail during the invocation itself. Can you verify that is the current behavior? I can fix that in hebo-gateway and include in v0.2.2.

turisanapo: Understand, should this be fixed here or on hebo ai gateway?

apps/gateway/src/index.ts
heiwen: Don't forget the trailing wildcard, to ensure that calls to individual models are working as well.

apps/gateway/src/services/model-resolver.ts
turisanapo: Updated

apps/gateway/src/gateway-config.ts
heiwen: Hebo Gateway. Would just be good if you could verify that it currently fails for you.

apps/gateway/src/services/model-resolver.ts
heiwen: 1/ 30 seconds feels extremely short. Suggest to start with 5 minutes for now.
2/ The TTL should only apply to the credentials, not the whole provider. Otherwise we re-initialize the auth as well.

apps/gateway/src/index.ts
turisanapo: Good one, updated

heiwen: `/models*` instead of `/models/*`

apps/gateway/src/services/model-resolver.ts
turisanapo: Done

apps/gateway/src/index.ts
```diff
@@ -15,24 +29,89 @@ import { errorHandler } from "./middlewares/error-handler";
 
 const PORT = Number(process.env.PORT ?? 3002);
 
+// Workaround for Elysia type issue with async handlers returning Response
+// See: https://github.com/elysiajs/elysia/issues/1721
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const asResponse = <T extends (...args: any[]) => Promise<Response>>(
```
heiwen: You should be able to just add an `as any` to each handler itself.

apps/gateway/src/index.ts
```diff
@@ -15,24 +29,89 @@ import { errorHandler } from "./middlewares/error-handler";
...
+      {
+        response: {
+          200: ModelListSchema,
+          400: OpenAIErrorSchema,
+          500: OpenAIErrorSchema,
+        },
+      },
+    )
+    .get(
+      "/v1/models/:modelId",
```
heiwen: Does this actually work for modelIds like "openai/gpt-oss-20b"?

apps/gateway/src/index.ts
```diff
@@ -15,24 +29,89 @@ import { errorHandler } from "./middlewares/error-handler";
...
+        .post(
+          "/chat/completions",
+          asResponse(({ request, dbClient }) =>
+            gw.handler(request, { dbClient }),
+          ),
+          {
+            parse: "none",
+            body: ChatCompletionsBodySchema,
+            response: {
+              200: ChatCompletionsSchema,
```
heiwen: How about `Streaming` responses?

apps/gateway/src/index.ts
turisanapo: Yes, it can be simplified.

apps/gateway/src/index.ts
turisanapo: Good catch, updated

apps/gateway/src/index.ts
turisanapo: Good catch, updated

apps/gateway/src/services/model-resolver.ts
```diff
@@ -11,11 +11,15 @@ import { injectMetadataCredentials } from "./aws-wif";
...
-  ResolveModelHookContext,
-  ResolveProviderHookContext,
+
+import {
+  CANONICAL_MODEL_IDS,
+  type ResolveModelHookContext,
+  type ResolveProviderHookContext,
 } from "@hebo-ai/gateway";
 
+const canonicalModelIds = new Set<string>(CANONICAL_MODEL_IDS);
```
heiwen, line 21: Why do you need to transform this into a set, can't you just use the original array?

turisanapo: Performance-wise, a set instead of an array allows O(1) lookup. Array complexity is O(n) since it must iterate through elements until a match is found.

apps/api/src/index.ts
```diff
@@ -25,12 +26,23 @@ const createApi = () =>
...
+          components: {
+            securitySchemes: {
+              bearerAuth: {
+                type: "http",
+                scheme: "bearer",
+                description: "API key or access token",
+              },
+            },
+          },
+          security: [{ bearerAuth: [] }],
```
heiwen: DRY: Feels like this should be extracted out to `openaiConfig`, the same as `corsConfig` and `getOtelConfig`.

apps/gateway/src/index.ts
```diff
@@ -25,26 +40,88 @@ export const createGateway = () =>
...
+          400: OpenAIErrorSchema,
+          500: OpenAIErrorSchema,
+        },
       },
-      (app) => app.use(completions).use(embeddings),
     )
-    // Public routes (no authentication required)
-    .group("/v1", (app) => app.use(models));
+    .get(
+      "/v1/models*",
```
heiwen: DRY: Shouldn't "/v1/models*" also capture "/v1/models" since `*` usually means "zero or more"?

apps/gateway/src/services/model-resolver.ts
```diff
@@ -0,0 +1,129 @@
...
+
+const configCache = new LRUCache<string, string>({
+  max: 100,
+  ttl: 5 * 60 * 1000, // 5 minutes
+});
+
+const providerCache = new LRUCache<string, ProviderV3>({
+  max: 100,
+});
```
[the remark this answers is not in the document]
turisanapo: This is expected

apps/api/src/index.ts
turisanapo: Good one, updated

apps/gateway/src/index.ts
turisanapo: Kept separated, updated url path.

apps/gateway/src/services/model-resolver.ts
```diff
@@ -0,0 +1,128 @@
...
+    const providerCacheKey = `${configCacheKey}:${configHash}`;
+    let provider = providerCache.get(providerCacheKey);
+
+    if (!provider) {
+      provider = createProvider(customProviderSlug, config.value);
+      providerCache.set(providerCacheKey, provider!);
+    }
+
+    return provider;
+  }
```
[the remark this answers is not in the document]
turisanapo: addressed

PR #172, opened 2026-02-07
Turisanapo/heb 305 mcp add world time tool

turisanapo:
## Summary by CodeRabbit

* **New Features**
  * Added a world-time tool to fetch current times across multiple time zones with locale-aware formatting, weekend indicators, and business-hours detection.

* **Chores**
  * Bumped app package version to 0.2.0 and updated the Model Context Protocol SDK dependency.


apps/mcp/src/aikit/current-timestamp.ts
```diff
@@ -0,0 +1,31 @@
...
+        .describe("IANA timezone (e.g., 'America/New_York'). Defaults to UTC."),
+    },
+  },
+  handler: async ({ timezone = "UTC" }: { timezone?: string }) => {
+    const now = new Date();
+    const formatted = now.toLocaleString("en-US", {
+      timeZone: timezone,
+      dateStyle: "full",
+      timeStyle: "long",
+    });
```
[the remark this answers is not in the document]
turisanapo: I tested with `NotATimezone` and the error is handled with a valid response from the server. It will let LLMs retry with no issues.

apps/mcp/src/aikit/world-time.ts
```diff
@@ -0,0 +1,103 @@
...
+        ),
+      options: z
+        .object({
+          locale: z
+            .string()
+            .optional()
+            .describe("Locale for formatting (e.g., 'en-US')"),
+          timeStyle: z.enum(["short", "medium", "long", "full"]).optional(),
+          dateStyle: z.enum(["short", "medium", "long", "full"]).optional(),
+          businessHours: z
```
heiwen: The `businessHours` functionality feels very edge case, suggest to remove.

apps/mcp/src/aikit/world-time.ts
```diff
@@ -0,0 +1,103 @@
...
+  // Partial match (e.g., "tokyo" → "Asia/Tokyo")
+  return IANA_ZONES.find((tz) => tz.toLowerCase().includes(q));
+};
+
+export const worldTimeTool = {
+  name: "world_time",
+  config: {
+    description:
+      "Get current time for one or more locations. Supports IANA timezones (e.g., 'America/New_York') or partial city names (e.g., 'tokyo' → 'Asia/Tokyo').",
+    inputSchema: {
```
heiwen, line 53: I liked the suggestion from ChatGPT to allow `queries`, `at` and `hints` which makes it much more powerful:

Image

turisanapo: Added `at` for querying specific points in time.

I would skip `hints` because the tool uses `Intl.supportedValuesOf("timeZone")`, which only provides IANA identifiers. There's no geodata, country mapping, or city database to resolve coordinates or disambiguate locations.

That intelligence should come from the AI layer above the tool. The LLM can interpret "Paris" in context and pass the correct IANA zone (`Europe/Paris` vs `America/Chicago`) to the tool. I would keep the tool simple and let the AI handle disambiguation, which I find a better approach.

apps/mcp/src/aikit/world-time.ts
turisanapo: Removed

apps/mcp/src/aikit/world-time.ts
heiwen: The hints are super important for the API, otherwise it becomes quite weak. LLMs are inherently bad at mapping data, they will hallucinate. There are libraries that provide the mappings. Just let your Cursor implement it.

turisanapo: Updated

turisanapo:
@claude can you help review this PR?

PR #180, opened 2026-02-16
feat(otel): enhance otel configuration

### LINKED ISSUE #179: Observability: reduce the generated spans

turisanapo:
This solves #179 and #182 

## Release Notes

* **New Features**
  * Centralized OpenTelemetry-based logging across all services with production-ready trace and metric exporters
  * Intelligent request filtering that enables tracing for non-GET requests and GET requests to non-root paths

* **Improvements**
  * Standardized logging initialization with simplified configuration
  * Enhanced observability through integrated structured logging and telemetry

* **Dependencies**
  * Expanded OpenTelemetry component support for comprehensive logging and metrics


packages/shared-api/lib/otel.ts
```diff
@@ -30,13 +30,23 @@ const getTraceExporterConfig = () => {
 const traceExporterConfig = getTraceExporterConfig();
 
 registerInstrumentations({
-  instrumentations: [new PrismaInstrumentation()],
+  instrumentations: [
+    new PrismaInstrumentation({
+      ignoreSpanTypes: ["prisma:client:compile", "prisma:client:serialize"],
```
heiwen: The only one that matters is "prisma:operation".

Image

packages/shared-api/lib/otel.ts
```diff
@@ -30,13 +30,23 @@ const getTraceExporterConfig = () => {
...
 });
 
 export const getOtelConfig = (
   serviceName: string,
 ): ElysiaOpenTelemetryOptions => ({
   serviceName,
+  checkIfShouldTrace: (request) => {
+    if (request.method !== "GET") return true;
+
+    const pathname = new URL(request.url).pathname;
```
heiwen: Perf: Creating a new URL is expensive, try to avoid.

packages/shared-api/lib/otel.ts
turisanapo: Updated to only track `prisma:client:operation` spans.

packages/shared-api/lib/otel.ts
turisanapo: Updated to use a simple string-based matching method.

packages/shared-api/lib/otel.ts
```diff
@@ -30,13 +31,21 @@ const getTraceExporterConfig = () => {
 const traceExporterConfig = getTraceExporterConfig();
 
 registerInstrumentations({
-  instrumentations: [new PrismaInstrumentation()],
+  instrumentations: [
+    new PrismaInstrumentation({
+      ignoreSpanTypes: [/^(?!prisma:client:operation$).+/],
```
heiwen: Perf: avoid regex

packages/shared-api/utils/url.ts
heiwen, line 1: Faster:

```
export function isRootPathUrl(url: string): boolean {
  const slash = url.indexOf("/");
  if (slash === -1) return false;

  const next = slash + 1;
  if (next >= url.length) return true;

  const c = url.charCodeAt(next);
  return c === 63 || c === 35;
}
```

apps/gateway/src/gateway-config.ts
```diff
@@ -5,6 +5,8 @@ import { gptOss20b, gptOss120b } from "@hebo-ai/gateway/models/openai";
 import { voyage35 } from "@hebo-ai/gateway/models/voyage";
 import { trace } from "@opentelemetry/api";
 
+import { createGatewayLogger } from "@hebo/shared-api/lib/logger";
```
heiwen: If this logger is specific to `gateway`, it shouldn't live in `shared-api`. But should it be specific to gateway or be a generic `otel` adapter?

packages/shared-api/lib/logger.ts
```diff
@@ -0,0 +1,114 @@
...
+  BatchLogRecordProcessor,
+  ConsoleLogRecordExporter,
+  LoggerProvider,
+} from "@opentelemetry/sdk-logs";
+
+import { logLevel } from "../env";
+
+import type { LogAttributes, Logger } from "@opentelemetry/api-logs";
+
+const enabledByLevel = {
```
heiwen: Use number mappings to simplify this:

```
const LEVEL = {
  trace: 5,
  debug: 10,
  info: 20,
  warn: 30,
  error: 40,
  silent: 50,
};
```

packages/shared-api/lib/logger.ts
```diff
@@ -0,0 +1,114 @@
...
+  otelLogger: Logger,
+  severityNumber: SeverityNumber,
+): ((...args: unknown[]) => void) => {
+  return (...args: unknown[]) => {
+    const first = args[0];
+    const second = args[1];
+    let body = "service log";
+    let attributes: LogAttributes | undefined;
+
+    if (typeof first === "string") {
```
heiwen: This mapping doesn't sound exactly correct.

The pino logger function signature is as follows:

```
export type LogFn = {
  (msg: string): void;
  (obj: Record<string, unknown>, msg?: string): void;
  (err: Error, msg?: string): void;
};
```

The mapping to otel `body`, `attributes` would be mostly to `body`. Attributes is for the `traceId` and the like:

**(msg: string):**
```
log("upstream failed");

otelLogger.emit({
  body: "upstream failed",
});
```

**(obj: Record<string, unknown>, msg?: string)**
```
log({ status: 502, model }, "upstream failed");

otelLogger.emit({
  body: {
    msg: "upstream failed",
    ...obj,
  },
});
```

**(err: Error, msg?: string)**
```
log(err, "request failed");

otelLogger.emit({
  severityNumber: SeverityNumber.ERROR,

  body: {
    msg,
    message: err.message,
    stack: err.stack,
  },

  attributes: {
    "error.type": err.name,
  },
});
```

packages/shared-api/lib/logger.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    otelLogger.emit({ severityNumber, body, attributes });
+  };
+};
+
+export const createGatewayLogger = (serviceName: string) => {
+  const normalizedServiceName =
+    serviceName.trim().length > 0 ? serviceName : "unknown_service:bun";
+
+  if (configuredLogLevel === "silent") {
```
heiwen: You can just not create a `LoggerProvider`, that has the same effect.

packages/shared-api/lib/logger.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  const loggerProvider = new LoggerProvider({
+    resource: resourceFromAttributes({
+      "service.name": normalizedServiceName,
+    }),
+    processors: [
+      new BatchLogRecordProcessor(new ConsoleLogRecordExporter(), {
+        maxQueueSize: 2048,
+        maxExportBatchSize: 512,
+        scheduledDelayMillis: 1000,
```
heiwen: Do you need to provide these defaults, are they different from the normal defaults?

packages/shared-api/lib/logger.ts
```diff
@@ -0,0 +1,114 @@
...
+      }),
+    ],
+  });
+  const otelLogger = loggerProvider.getLogger(normalizedServiceName);
+
+  return {
+    trace: enabledLevels.trace
+      ? createLogMethod(otelLogger, SeverityNumber.TRACE)
```
heiwen: Feels quite heavy to recreate the whole logic for every method. The only thing that is different is the log level. Simplify the code towards that.

packages/shared-api/lib/otel.ts
```diff
@@ -29,11 +35,18 @@ const getTraceExporterConfig = () => {
...
+  : new SimpleSpanProcessor(new ConsoleSpanExporter());
 
 registerInstrumentations({
   instrumentations: [
     new PrismaInstrumentation({
-      ignoreSpanTypes: [/^(?!prisma:client:operation$).+/],
+      ignoreSpanTypes: [
+        "prisma:client:compile",
+        "prisma:client:serialize",
+        "prisma:client:db_query",
```
heiwen: There seems to be more that we should ignore: https://www.prisma.io/docs/orm/prisma-client/observability-and-logging/opentelemetry-tracing#trace-output

packages/shared-api/lib/logger.ts
heiwen: There are a few additional cases that need to be handled for pino compatibility:

1/ If `obj` has an `err` property, that is the same as a top-level type `Error`
2/ Errors in TS are not iterable / enumerable, meaning you loose all the details, unless you manually serialize them. I used a small helper library for that `serialize-error`.

Have a look at how I solved that in the default console logger:
https://github.com/8monkey-ai/hebo-gateway/blob/main/src/logger/default.ts

you can probably reuse most of that code.

apps/gateway/src/index.ts
```diff
@@ -112,6 +113,18 @@ export const createGateway = () =>
...
+    await shutdownLoggers();
+    app.stop();
+  };
+
+  process.once("SIGINT", () => {
+    void shutdown();
+  });
+  process.once("SIGTERM", () => {
+    void shutdown();
+  });
```
heiwen: Is this really necessary?

packages/shared-api/lib/logger.ts
```diff
@@ -1,114 +1,172 @@
...
-      body = second;
+      otelLogger.emit({
+        severityNumber,
+        body: first,
+        attributes:
+          second && typeof second === "object"
+            ? (second as LogAttributes)
+            : undefined,
+      });
+      return;
```
heiwen: Code is easier to read / maintain if you collect all attributes first and then do a single `emit()`.

packages/shared-api/lib/logger.ts
```diff
@@ -1,114 +1,172 @@
...
-      ? createLogMethod(otelLogger, SeverityNumber.WARN)
-      : noop,
-    error: enabledLevels.error
-      ? createLogMethod(otelLogger, SeverityNumber.ERROR)
-      : noop,
+    trace: createLogMethod(otelLogger, "trace"),
+    debug: createLogMethod(otelLogger, "debug"),
+    info: createLogMethod(otelLogger, "info"),
+    warn: createLogMethod(otelLogger, "warn"),
+    error: createLogMethod(otelLogger, "error"),
```
heiwen: Avoid creating logger methods if the log level is disabled.

packages/shared-api/lib/logger.ts
```diff
@@ -1,114 +1,172 @@
...
+  return {
+    url: betterStackConfig.endpoint,
+    metadata,
+    compression: CompressionAlgorithm.GZIP,
+  };
 };
-const enabledLevels = getEnabledLevels(configuredLogLevel);
 
-const noop = () => {};
+const getOrCreateOtelLogger = (serviceName: string) => {
```
heiwen: Why would this ever need a cache, shouldn't this just be called once on startup?

packages/shared-api/lib/logger.ts
```diff
@@ -1,114 +1,172 @@
...
+  const cachedLogger = loggerByServiceName.get(serviceName);
+  if (cachedLogger) return cachedLogger;
+
+  const logExporterConfig = getLogExporterConfig();
+  const logRecordProcessor = logExporterConfig
+    ? new BatchLogRecordProcessor(new OTLPLogExporter(logExporterConfig), {
+        maxQueueSize: 2048,
+        maxExportBatchSize: 512,
+        scheduledDelayMillis: 1000,
```
heiwen: Are these the default values? If yes, remove, if no, why different from default?

packages/shared-api/lib/otel.ts
turisanapo: Updated.

packages/shared-api/utils/url.ts
turisanapo: Updated.

apps/gateway/src/gateway-config.ts
turisanapo: Updated.

packages/shared-api/lib/logger.ts
turisanapo: Updated.

packages/shared-api/lib/logger.ts
turisanapo: Good catch, fixed.

packages/shared-api/lib/logger.ts
turisanapo: Handled now before the log handler is created.

packages/shared-api/lib/logger.ts
turisanapo: Removed

packages/shared-api/lib/logger.ts
turisanapo: Just creating the handler once now.

packages/shared-api/lib/otel.ts
turisanapo: Updated the list.

apps/gateway/src/index.ts
turisanapo: Since we are exporting logs in batches, there may be cases where the application is shut down and not all logs have been exported.

However, I realised that the impact seems to be minimal. I currently removed it, until we don't realize it is an actual issue to be addressed.

packages/shared-api/lib/logger.ts
turisanapo: Done.

packages/shared-api/lib/logger.ts
turisanapo: Done.

packages/shared-api/lib/logger.ts
turisanapo: Technically, at startup, the service may be init multiple times, resulting in duplicated exports. I currently remove it until we actually observe the issue.

packages/shared-api/lib/logger.ts
turisanapo: Removed

apps/gateway/src/index.ts
heiwen: If that case exists, it would also exist for the normal span exports?

packages/shared-api/lib/logger.ts
heiwen: That would be the case for any other config as well though, wouldn't it?

apps/gateway/src/gateway-config.ts
```diff
@@ -59,10 +58,14 @@ export const gw = gateway({
...
-  logger: pino({ level: logLevel }),
+  logger: createLogger("hebo-ai-gateway"),
   telemetry: {
     enabled: true,
     tracer: trace.getTracer("hebo-ai-gateway"),
-    attributes: "full",
+    signals: {
+      gen_ai: "full",
+      http: "recommended",
+      hebo: "recommended",
```
heiwen: Suggest to go with full on hebo, that provides debugging information that might be helpful while we are still stabilizing it.

apps/api/src/index.ts
```diff
@@ -21,7 +19,6 @@ const API_URL = process.env.API_URL ?? `http://localhost:${PORT}`;
 const createApi = () =>
   new Elysia()
     .use(opentelemetry(getOtelConfig("hebo-api")))
-    .use(logger(getLoggerOptions()))
```
heiwen, line 24: Without a logger, how does normal development mode currently surface:
1/ exceptions
2/ access logs

apps/auth/src/index.ts
```diff
@@ -14,7 +12,6 @@ const PORT = Number(process.env.PORT ?? 3000);
 const createAuth = () =>
   new Elysia()
     .use(opentelemetry(getOtelConfig("hebo-auth")))
-    .use(logger(getLoggerOptions()))
```
heiwen, line 17: Without a logger, how does normal development mode currently surface:
1/ exceptions
2/ access logs

apps/mcp/src/index.ts
```diff
@@ -25,7 +23,6 @@ function createMcpServer() {
 const createMcp = () =>
   new Elysia()
     .use(opentelemetry(getOtelConfig("hebo-mcp")))
-    .use(logger(getLoggerOptions()))
```
heiwen, line 28: Without a logger, how does normal development mode currently surface:
1/ exceptions
2/ access logs

packages/shared-api/lib/logger.ts
heiwen, line 1: Feels this is really two files:
1/ A general pino compatible otel logger
2/ Our specific configuration of that

Suggest to separate out.

packages/shared-api/lib/logger.ts
```diff
@@ -1,22 +1,189 @@
...
+  info: 30,
+  warn: 40,
+  error: 50,
+  silent: 60,
+};
+
+const getLogLevelWeight = (level: string): number =>
+  levelWeightByLevel[level as ConfigLogLevel] ?? levelWeightByLevel.info;
+
+const getSeverityNumber = (level: LogLevel): SeverityNumber => {
```
heiwen: 1/ Naming: `getOtelSeverityNumber`
2/ Style: opening & closing brackets are not necessary and make the code longer than it needs to be

packages/shared-api/lib/logger.ts
```diff
@@ -1,22 +1,189 @@
...
+    case "warn": {
+      return SeverityNumber.WARN;
+    }
+    case "error": {
+      return SeverityNumber.ERROR;
+    }
+  }
+};
+
+const configuredLogLevelWeight = getLogLevelWeight(logLevel);
```
heiwen: Feels like an unnecessary wrapper?

packages/shared-api/lib/logger.ts
```diff
@@ -1,22 +1,189 @@
...
+
+const noop = () => {};
+
+const serializeError = (error: Error) => ({
```
heiwen: This misses the .cause object (and it's recursive children), which removes a lot of value-able information. That's why I use the 3rd party library.

packages/shared-api/lib/logger.ts
```diff
@@ -1,22 +1,189 @@
...
+const buildObjectLog = (
+  severityNumber: SeverityNumber,
+  obj: Record<string, unknown>,
+  msg?: string,
+): Parameters<Logger["emit"]>[0] => {
+  const err = obj.err;
+  if (err instanceof Error) {
+    const rest = { ...obj };
+    delete rest.err;
+    return {
```
heiwen: As mentioned in previous comment, prefer building the object once with a single return, makes the code easier to reason about.

packages/shared-api/lib/logger.ts
```diff
@@ -1,22 +1,189 @@
...
+
+const serializeError = (error: Error) => ({
+  message: error.message,
+  stack: error.stack,
+});
+
+const asBody = (value: unknown) =>
+  value as NonNullable<Parameters<Logger["emit"]>[0]["body"]>;
+
+const buildObjectLog = (
```
heiwen: This function is only called a single time. Inline it for better clarity of the control flow.

packages/shared-api/lib/logger.ts
```diff
@@ -1,22 +1,189 @@
...
+    } else if (first instanceof Error) {
+      logRecord = {
+        severityNumber,
+        body: asBody({
+          ...(msg ? { msg } : {}),
+          ...serializeError(first),
+        }),
+        attributes: { "error.type": first.name },
+      };
+    } else if (typeof first === "object" && first !== null) {
```
heiwen: If first is an "object", it can't be "null", can it?

packages/shared-api/lib/logger.ts
```diff
@@ -1,22 +1,189 @@
...
+      info: noop,
+      warn: noop,
+      error: noop,
+    };
+  }
+
+  const otelLogger = createOtelLogger(serviceName || "unknown_service:bun");
+  const log = createLogHandler(otelLogger);
+
+  return {
```
heiwen: 1/ You can use a `for` loop over the `LogLevel`'s to simplify this.
2/ It should only assign `log()` if the `LogLevel` has been enabled

See my version in gateway for inspiration.

packages/shared-api/lib/logger.ts
```diff
@@ -1,22 +1,189 @@
...
-    trace_flags: spanContext.traceFlags.toString(16).padStart(2, "0"),
+    url: betterStackConfig.endpoint,
+    metadata,
+    compression: CompressionAlgorithm.GZIP,
   };
 };
 
-export const getLoggerOptions = () => {
-  return { level: logLevel, customProps };
+const createOtelLogger = (serviceName: string) => {
```
heiwen: Feels this code is closer to what otel.ts does, should probably live there.

packages/shared-api/middlewares/auth.ts
```diff
@@ -74,8 +71,6 @@ export const authService = new Elysia({ name: "auth-service" })
     }
 
     if (error || !session) {
-      log.info({ error }, "Authentication failed or no credentials provided");
```
heiwen, line 77: How would I know in development about this error condition?

packages/shared-api/utils/url.ts
heiwen: Don't see it updated.

turisanapo: This is the commit for the update: https://github.com/8monkey-ai/hebo-platform/pull/180/changes/22e12f5e12f781ea98d142dca669bf0e64a49415

I slightly changed the one you proposed because manual tests were wrongly returning false for the following 2 URLs:

- `https://api.example.com`
- `https://api.example.com/`

apps/gateway/src/index.ts
turisanapo: Yes, correct, if it may happen for both. However, I would address this only once we realise (i) the issue actually happens for our setup, (ii) we want to maximise the reliability of the observability stack.

packages/shared-api/lib/logger.ts
turisanapo: Correct. Let's address it only once we realise we actually have duplicated exports.

apps/api/src/index.ts
turisanapo: By “access logs” I assume you mean the usual per-request HTTP logs (method, route, status, latency).

In dev they are both surfaced through span console output:

1/ Exceptions are visible in the request trace/span (status and details)
2/ Access logs are represented by the request span metadata instead of separate logger lines

packages/shared-api/lib/logger.ts
turisanapo: Updated

packages/shared-api/lib/logger.ts
turisanapo: This is on purpose: it raises `createLogHandler` cognitive complexity above our Sonar threshold.

packages/shared-api/lib/logger.ts
turisanapo: In JS, `typeof null` is `object`.

packages/shared-api/middlewares/auth.ts
turisanapo: The information is surfaced by the response span, which includes an HTTP status of `401`.

apps/gateway/src/gateway-config.ts
turisanapo: Done

apps/auth/src/index.ts
turisanapo: In dev they are both surfaced through span console output:

1/ Exceptions are visible in the request trace/span (status and details)
2/ Access logs are represented by the request span metadata instead of separate logger lines

apps/mcp/src/index.ts
turisanapo: In dev they are both surfaced through span console output:

1/ Exceptions are visible in the request trace/span (status and details)
2/ Access logs are represented by the request span metadata instead of separate logger lines

packages/shared-api/lib/logger.ts
turisanapo: Done

packages/shared-api/lib/logger.ts
turisanapo: Done

packages/shared-api/lib/logger.ts
turisanapo: Updated

packages/shared-api/lib/logger.ts
turisanapo: Done

packages/shared-api/lib/logger.ts
turisanapo: Done

packages/shared-api/lib/logger.ts
turisanapo: Removed logger module

apps/api/src/index.ts
```diff
@@ -18,7 +18,7 @@ const API_URL = process.env.API_URL ?? `http://localhost:${PORT}`;
 
 const createApi = () =>
   new Elysia()
-    .use(opentelemetry(getOtelConfig("hebo-api")))
+    .use(opentelemetry(getOtelTraceConfig("hebo-api")))
```
heiwen: How about `metrics`, does this config capture these as well?

apps/gateway/src/gateway-config.ts
```diff
@@ -58,14 +60,18 @@ export const gw = gateway({
     resolveModelId,
     resolveProvider,
   },
-  logger: createLogger("hebo-ai-gateway"),
+  logger: createPinoCompatibleOtelLogger({
```
heiwen: Feels the configuration interface should be similar to the "core" otel configuration.

packages/shared-api/middlewares/auth.ts
heiwen: My understanding is that in development you don't want to rely on otel spans.

packages/shared-api/lib/logger.ts
heiwen: Lint rules are just "best practices", not "hard rules". I can tell you, having it as a separate function did definitely increase my cognitive overload. Inlining would reduce it.

packages/shared-api/lib/otel-pino-adapter.ts
```diff
@@ -0,0 +1,148 @@
...
+  warn: 40,
+  error: 50,
+  silent: 60,
+};
+
+const getLogLevelWeight = (level: string): number =>
+  levelWeightByLevel[level as ConfigLogLevel] ?? levelWeightByLevel.info;
+
+const getOtelSeverityNumber = (level: LogLevel): SeverityNumber => {
+  switch (level) {
```
heiwen: Style: opening & closing brackets are not necessary and make the code longer than it needs to be

apps/mcp/src/index.ts
heiwen: My understanding is that in development you don't want to rely on otel spans.

apps/auth/src/index.ts
heiwen: My understanding is that in development you don't want to rely on otel spans.

apps/api/src/index.ts
heiwen: My understanding is that in development you don't want to rely on otel spans.

apps/gateway/src/index.ts
heiwen: Aligned

packages/shared-api/lib/otel-pino-adapter.ts
```diff
@@ -0,0 +1,148 @@
...
+      };
+    }
+    otelLogger.emit(logRecord);
+  };
+};
+
+export const createPinoCompatibleOtelLogger = ({
+  serviceName,
+  logLevel,
+  createOtelLogger,
```
heiwen: This doesn't feel like a clean separation of concerns. Shouldn't need to pass a "create" function, but just a `Logger` instead. Does the `Logger` have config for level's already?

packages/shared-api/lib/otel.ts
```diff
@@ -57,13 +77,17 @@ registerInstrumentations({
...
   serviceName: string,
 ): ElysiaOpenTelemetryOptions => ({
   serviceName,
   checkIfShouldTrace: (request) => {
     if (request.method !== "GET") return true;
     return !isRootPathUrl(request.url);
   },
-  spanProcessors: [spanProcessor],
+  spanProcessors: [
+    otlpExporterConfig
```
heiwen: Should this be based on NODE_ENV for clarity?

apps/api/src/index.ts
turisanapo: Based on my understanding, it requires additional explicit configuration.

Shall we include it now or when we start emitting metrics?

packages/shared-api/lib/otel-pino-adapter.ts
heiwen, line 1: This would be `utils` not `lib` (no domain logic).

apps/api/src/index.ts
heiwen: Gateway is emitting metrics for duration, token usage and memory consumption.

packages/shared-api/utils/url.ts
heiwen: Minimal fix without introducing a second `indexOf`:

```
export function isRootPathUrl(url: string): boolean {
  const slash = url.indexOf("/", 8);
  if (slash === -1) return true;

  const next = slash + 1;
  if (next >= url.length) return true;

  const c = url.charCodeAt(next);
  return c === 63 || c === 35;
}
```

heiwen:
I wasn't happy with how `serialize-error` library handles `ZodErrors`, so I implemented my own version:

```
function serializeError(err: unknown, _seen?: WeakSet<object>): Record<string, unknown> {
  if (!(err instanceof Error)) return { message: String(err) };

  const seen = _seen ?? new WeakSet();
  if (seen.has(err)) return { name: err.name, message: err.message, circular: true };
  seen.add(err);

  const out: Record<string, unknown> = {};

  for (const k of Object.getOwnPropertyNames(err)) {
    if (k.startsWith("_")) continue;

    let val: unknown;
    try {
      val = (err as any)[k];
    } catch {
      val = "[Unreadable]";
    }

    if (typeof val === "bigint") val = `${val}n`;

    // FUTURE: check for circular references within val
    out[String(k)] = val instanceof Error ? serializeError(val, seen) : val;
  }

  return out;
}
```

packages/shared-api/utils/url.ts
turisanapo: The `8` assumes `https://` and would misidentify the path start for `http://` URLs (or any other scheme length).

apps/api/src/index.ts
turisanapo: Introduced logger.

apps/auth/src/index.ts
turisanapo: Introduced logger.

apps/mcp/src/index.ts
turisanapo: Introduced logger.

packages/shared-api/lib/logger.ts
turisanapo: Updated.

packages/shared-api/middlewares/auth.ts
turisanapo: Introduced logger.

apps/gateway/src/gateway-config.ts
turisanapo: Simplified.

packages/shared-api/lib/otel-pino-adapter.ts
turisanapo: Simplified.

packages/shared-api/lib/otel-pino-adapter.ts
turisanapo: Simplified.

packages/shared-api/lib/otel.ts
turisanapo: Updated.

packages/shared-api/lib/otel-pino-adapter.ts
turisanapo: Updated.

turisanapo:
> I wasn't happy with how `serialize-error` library handles `ZodErrors`, so I implemented my own version:
> 
> ```
> function serializeError(err: unknown, _seen?: WeakSet<object>): Record<string, unknown> {
>   if (!(err instanceof Error)) return { message: String(err) };
> 
>   const seen = _seen ?? new WeakSet();
>   if (seen.has(err)) return { name: err.name, message: err.message, circular: true };
>   seen.add(err);
> 
>   const out: Record<string, unknown> = {};
> 
>   for (const k of Object.getOwnPropertyNames(err)) {
>     if (k.startsWith("_")) continue;
> 
>     let val: unknown;
>     try {
>       val = (err as any)[k];
>     } catch {
>       val = "[Unreadable]";
>     }
> 
>     if (typeof val === "bigint") val = `${val}n`;
> 
>     // FUTURE: check for circular references within val
>     out[String(k)] = val instanceof Error ? serializeError(val, seen) : val;
>   }
> 
>   return out;
> }
> ```

Thanks.

apps/api/src/index.ts
turisanapo: Introduced configuration for metrics.

packages/shared-api/middlewares/auth.ts
heiwen: You need to restore this line then?

packages/shared-api/utils/url.ts
heiwen: It works correctly for all of them:

```
> mod.isRootPathUrl("https://api.example.com")
true
> mod.isRootPathUrl("https://api.example.com/")
true
> mod.isRootPathUrl("http://api.example.com")
true
> mod.isRootPathUrl("http://api.example.com/")
true
> mod.isRootPathUrl("api.example.com")
true
> mod.isRootPathUrl("api.example.com/")
true
> mod.isRootPathUrl("http://api.example.com/a")
false
> mod.isRootPathUrl("https://api.example.com/a")
false
> mod.isRootPathUrl("api.example.com/a")
false
```

apps/api/src/index.ts
```diff
@@ -21,7 +20,7 @@ const API_URL = process.env.API_URL ?? `http://localhost:${PORT}`;
 const createApi = () =>
   new Elysia()
     .use(opentelemetry(getOtelConfig("hebo-api")))
-    .use(logger(getLoggerOptions()))
+    .use(serviceLogging("hebo-api"))
```
heiwen: `serviceLogging` is a weird name, why not just `logger`?

packages/shared-api/lib/otel.ts
```diff
@@ -27,17 +42,68 @@ const getTraceExporterConfig = () => {
   };
 };
 
-const traceExporterConfig = getTraceExporterConfig();
+const otlpExporterConfig = isProduction
```
heiwen: 1/ Readability: Why not inline this with all the rest of the `createOtelLogger`?
2/ Once it is inlined, it becomes clear that isProduction is checked twice and unnecessary.

packages/shared-api/lib/otel.ts
```diff
@@ -27,17 +42,68 @@ const getTraceExporterConfig = () => {
...
+  ? getOtlpGrpcExporterConfig()
+  : undefined;
+
+export const createOtelLogger = (serviceName: string, logLevel: string) => {
```
heiwen: Naming: Why do you use `createOtelLogger` here and `getOtelConfig` in the other place?

packages/shared-api/middlewares/auth.ts
```diff
@@ -74,9 +74,11 @@ export const authService = new Elysia({ name: "auth-service" })
...
+      logger.info(
+        { error },
+        "Authentication failed or no credentials provided",
+      );
 
-      // Clear the session cookie when unauthorized
```
heiwen, line 79: Suggest to keep comment. Not clear what this is doing without that.

packages/shared-api/middlewares/logging.ts
```diff
@@ -0,0 +1,41 @@
...
+  logger = createPinoCompatibleOtelLogger(
+    createOtelLogger(serviceName, logLevel),
+  ),
+) =>
+  new Elysia({
+    name: "hebo-logging",
+    seed: { serviceName },
+  })
+    .decorate("logger", logger)
+    .onRequest(function logIncomingRequest({ request }) {
```
heiwen: Do you need a named function here?

packages/shared-api/middlewares/logging.ts
```diff
@@ -0,0 +1,41 @@
...
+      if (isProduction) return;
```
heiwen: Can you make the whole handler conditional of production, so that it won't show up in traces in production?

packages/shared-api/utils/otel/log-levels.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const otelLogLevels = [
+  "trace",
+  "debug",
+  "info",
+  "warn",
+  "error",
+] as const;
+
+export type OtelLogLevel = (typeof otelLogLevels)[number];
```
heiwen: This is never used externally.

packages/shared-api/utils/otel/log-levels.ts
```diff
@@ -0,0 +1,19 @@
+import { SeverityNumber } from "@opentelemetry/api-logs";
+
+export const otelLogLevels = [
```
heiwen: Do you need this separate array, or can you just infer the log levels from the keys of `otelServerityByLevel` in external consumers.

packages/shared-api/utils/otel/pino-adapter.ts
```diff
@@ -0,0 +1,109 @@
...
+    if (typeof val === "bigint") val = `${val}n`;
+
+    out[String(k)] = val instanceof Error ? serializeError(val, seen) : val;
+  }
+
+  return out;
+};
+
+type LogLevel = (typeof otelLogLevels)[number];
+const getOtelSeverityNumber = (level: LogLevel) => otelSeverityByLevel[level];
```
heiwen: Unnecessary wrapper

packages/shared-api/utils/otel/pino-adapter.ts
```diff
@@ -0,0 +1,109 @@
...
+
+const createLogHandler = (
+  otelLogger: Logger,
+): ((level: LogLevel, ...args: unknown[]) => void) => {
+  return (level: LogLevel, ...args: unknown[]) => {
+    const severityNumber = getOtelSeverityNumber(level);
+
+    const first = args[0];
+    const second = args[1];
+    const msg =
```
heiwen: Weird flow, you start using second before first. Hard to read.

packages/shared-api/utils/otel/pino-adapter.ts
```diff
@@ -0,0 +1,109 @@
...
+      typeof second === "string" && second.length > 0 ? second : undefined;
+    let logRecord: Parameters<Logger["emit"]>[0];
+
+    if (typeof first === "string") {
+      logRecord = {
+        severityNumber,
```
heiwen: `severityNumber` is always the same, assign it when creating the logRecord to avoid duplication

packages/shared-api/utils/otel/pino-adapter.ts
```diff
@@ -0,0 +1,109 @@
...
+        body: asBody({
+          ...(msg ? { msg } : {}),
+          ...serializeError(first),
+        }),
+        attributes: { "error.type": first.name },
+      };
+    } else if (typeof first === "object" && first !== null) {
+      const obj = first as Record<string, unknown>;
```
heiwen: This cast shouldn't be necessary, since you already narrowed the type in the if statement.

packages/shared-api/utils/otel/pino-adapter.ts
```diff
@@ -0,0 +1,109 @@
...
+  return out;
+};
+
+type LogLevel = (typeof otelLogLevels)[number];
+const getOtelSeverityNumber = (level: LogLevel) => otelSeverityByLevel[level];
+
+const asBody = (value: unknown) =>
+  value as NonNullable<Parameters<Logger["emit"]>[0]["body"]>;
+
+const createLogHandler = (
```
heiwen: This overall flow is very hard to understand, I stopped half-way through.

Please have another look at the gateway version for clearer control flow:
https://github.com/8monkey-ai/hebo-gateway/blob/d994274855e1071fb89cfd14b96b64f291bd1d54/src/logger/default.ts

packages/shared-api/utils/otel/pino-adapter.ts
```diff
@@ -0,0 +1,109 @@
...
+        severityNumber,
+        body: first === undefined ? "service log" : String(first),
+      };
+    }
+    otelLogger.emit(logRecord);
+  };
+};
+
+export const createPinoCompatibleOtelLogger = (otelLogger: Logger) => {
+  const log = createLogHandler(otelLogger);
```
heiwen: Why is it called `createLogHandler`, it's not actually `creating` anything, it's just `handling`, isn't it?

packages/shared-api/utils/url.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  const queryStart = url.indexOf("?", pathStart);
+  const hashStart = url.indexOf("#", pathStart);
+  let pathEnd = url.length;
+  if (queryStart !== -1) pathEnd = Math.min(pathEnd, queryStart);
+  if (hashStart !== -1) pathEnd = Math.min(pathEnd, hashStart);
+
+  const path = url.slice(pathStart, pathEnd);
+  return path.length === 0 ? "/" : path;
+}
```
heiwen: 1/ I put this code in ChatGPT and asked "Do you see a way to make this faster".

This is what it came up with in one-shot:

```
export function getPathnameFromUrl(url: string): string {
  const start = getPathStartIndex(url);
  if (start === -1) return "/";

  let end = start;
  for (; end < url.length; end++) {
    const c = url.charCodeAt(end);
    if (c === 63 || c === 35) break; // '?' or '#'
  }

  if (end === start) return "/";
  return url.slice(start, end);
}
```

2/ If this really only runs in development and not in production, speed is not an issue, and you can just use `new URL(...).pathname` at the call site.

eslint.config.mjs
```diff
@@ -122,13 +122,20 @@ const eslintConfig = [
...
+      'sonarjs/cognitive-complexity': ['error', 15],
       'sonarjs/todo-tag': 'warn',
       'tailwindcss/no-custom-classname': 'off',
       '@typescript-eslint/no-explicit-any': 'warn',
       'turbo/no-undeclared-env-vars': 'warn',
       'unicorn/no-abusive-eslint-disable': 'warn'
     },
   },
+  {
+    files: ['packages/shared-api/utils/otel/pino-adapter.ts'],
```
heiwen: Use // eslint-disable-next-line at the call site

packages/shared-api/utils/url.ts
turisanapo: Updated

apps/api/src/index.ts
turisanapo: Updated

packages/shared-api/lib/otel.ts
turisanapo: Updated

packages/shared-api/lib/otel.ts
turisanapo: Made naming consistent

packages/shared-api/middlewares/auth.ts
turisanapo: Restored

packages/shared-api/middlewares/logging.ts
turisanapo: We still want to use the logger for errors.

packages/shared-api/middlewares/logging.ts
turisanapo: It is to show a named span. In production, we still want to use the logger for errors.

packages/shared-api/utils/otel/log-levels.ts
turisanapo: Removed.

packages/shared-api/utils/otel/log-levels.ts
turisanapo: Removed.

packages/shared-api/utils/otel/pino-adapter.ts
turisanapo: I refactored the entire module. Please have a look again.

packages/shared-api/utils/otel/pino-adapter.ts
turisanapo: I refactored the entire module. Please have a look again.

packages/shared-api/utils/otel/pino-adapter.ts
turisanapo: I refactored the entire module. Please have a look again.

packages/shared-api/utils/otel/pino-adapter.ts
turisanapo: I refactored the entire module. Please have a look again.

packages/shared-api/utils/otel/pino-adapter.ts
turisanapo: I refactored the entire module. Please have a look again. Thank you for the suggestion.

packages/shared-api/utils/otel/pino-adapter.ts
turisanapo: Renamed.

packages/shared-api/utils/url.ts
turisanapo: Updated, thanks for the suggestion.

eslint.config.mjs
turisanapo: Removed, after the refactor, there is no need to update/ignore the linter.

packages/shared-api/middlewares/logging.ts
heiwen: You can still make the onRequest handler conditional then, just leave the onError one.

packages/shared-api/middlewares/logging.ts
heiwen: But you don't care about the access logging for onRequest

apps/api/src/index.ts
```diff
@@ -20,8 +19,8 @@ const API_URL = process.env.API_URL ?? `http://localhost:${PORT}`;
 
 const createApi = () =>
   new Elysia()
-    .use(opentelemetry(getOtelConfig("hebo-api")))
-    .use(logger(getLoggerOptions()))
+    .use(opentelemetry(createOtelConfig("hebo-api")))
```
heiwen: We are not "creating" a config, the config has been created at development time. We are "getting" a config.

packages/shared-api/lib/otel.ts
```diff
@@ -27,17 +42,66 @@ const getTraceExporterConfig = () => {
...
+  const loggerProvider = new LoggerProvider({
+    resource: resourceFromAttributes({
+      "service.name": serviceName,
+    }),
+    loggerConfigurator: createLoggerConfigurator([
+      {
+        pattern: "*",
+        config: {
+          minimumSeverity:
+            otelSeverityByLevel[logLevel as keyof typeof otelSeverityByLevel] ??
```
heiwen: Should this be `debug` for development?

packages/shared-api/middlewares/auth.ts
```diff
@@ -10,6 +9,8 @@ import { AuthError, BadRequestError } from "../errors";
 import { betterAuthCookieOptions } from "../lib/cookie-options";
 import { getSecret } from "../utils/secrets";
 
+import type { ServiceLogger } from "./logging";
```
heiwen: Why is this called `ServiceLogger` instead of just `Logger`? Are you planning to use non service loggers in the same module?

packages/shared-api/utils/otel/log-levels.ts
heiwen, line 1: Feels like this should just be part of the pino-otel-adapter.

packages/shared-api/utils/otel-pino-adapter.ts
heiwen, line 1: Not really a `factory`, the previous `adapter` nomenclature was correct.

packages/shared-api/utils/otel/logger-factory.ts
```diff
@@ -0,0 +1,100 @@
...
+  let err: Record<string, unknown> | undefined;
+  let errorType: string | undefined;
+  let msg: string | undefined;
+
+  if (first instanceof Error) {
+    err = serializeError(first);
+    errorType = first.name;
+  } else if (isRecord(first)) {
+    if (first["err"] instanceof Error) {
+      errorType = first["err"].name;
```
heiwen: No need to store this in a separate const. Just check if err exists in line 91 and assign it there.

packages/shared-api/utils/otel/logger-factory.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  if (err && msg === undefined) {
+    msg = err["message"] as string;
+  }
+
+  const body: Record<string, unknown> = { ...obj };
+  if (msg) body.msg = msg;
+  if (err) body.err = err;
+
+  const record: LogRecord = { body: body as LogRecord["body"] };
```
heiwen: `as cast` should be avoided as they can introduce runtime bugs. It should not be necessary if you type `body` correctly in line 76.

packages/shared-api/utils/otel/logger-factory.ts
```diff
@@ -0,0 +1,100 @@
...
+  if (errorType) record.attributes = { "error.type": errorType };
+  return record;
+};
+
+const severityEntries = Object.entries(otelSeverityByLevel) as [
+  LogLevel,
+  (typeof otelSeverityByLevel)[LogLevel],
+][];
+
+export const createServiceLogger = (otelLogger: Logger) =>
```
heiwen: Why `ServiceLogger`. This module is in `util` meaning it implies zero domain logic. It is a general pino <> otel adapter?

packages/shared-api/utils/otel/logger-factory.ts
```diff
@@ -0,0 +1,100 @@
...
+  const body: Record<string, unknown> = { ...obj };
+  if (msg) body.msg = msg;
+  if (err) body.err = err;
+
+  const record: LogRecord = { body: body as LogRecord["body"] };
+  if (errorType) record.attributes = { "error.type": errorType };
+  return record;
+};
+
+const severityEntries = Object.entries(otelSeverityByLevel) as [
```
heiwen: Avoid the "as cast" by typing `otelSeverityByLevel` correctly.

packages/shared-api/middlewares/logging.ts
```diff
@@ -0,0 +1,35 @@
...
+  new Elysia({ name: "hebo-logging" })
+    .decorate("logger", logger)
+    .onRequest(function logIncomingRequest({ request }) {
+      if (isProduction) return;
+      logger.info(getRequestMeta(request), "request:incoming");
+    })
+    .onError(({ request, error, set }) => {
+      logger.error(
+        {
+          ...getRequestMeta(request),
```
heiwen: Is `getRequestMeta` really helpful, or do you already have that because (i) in development requests are ordered (ii) in production it's attached to a span.

packages/shared-api/lib/otel.ts
turisanapo: The logger is created using `logLevel`, which is `debug` by default in development, unless specified by the user using env variables.

The default to info here guards against mispelled or not supported manually configured log levels using env variables.

packages/shared-api/utils/otel/log-levels.ts
turisanapo: `otelSeverityByLevel` is currently used by both the logger adapter and the otel module. If we move this helper into the logger adapter module, the otel module will have to import it from there, which effectively makes otel depend on the logger adapter.

This would change the dependency direction and create the wrong inversion.

packages/shared-api/middlewares/logging.ts
turisanapo: Removed

packages/shared-api/utils/otel/logger-factory.ts
turisanapo: Replaced with simpler `LogLevel[]` since `Object.keys()` always returns `string[]`.

packages/shared-api/utils/otel/logger-factory.ts
turisanapo: Renamed

packages/shared-api/utils/otel/logger-factory.ts
turisanapo: Removed

packages/shared-api/utils/otel/logger-factory.ts
turisanapo: Removed

packages/shared-api/utils/otel-pino-adapter.ts
turisanapo: Renamed

packages/shared-api/middlewares/auth.ts
turisanapo: Renamed

apps/api/src/index.ts
turisanapo: Renamed

apps/gateway/src/gateway-config.ts
```diff
@@ -59,10 +60,14 @@ export const gw = gateway({
     resolveModelId,
     resolveProvider,
   },
-  logger: pino({ level: logLevel }),
+  logger: createPinoAdapter(getOtelLogger("hebo-ai-gateway", logLevel)),
```
heiwen: Naming: should read `createPinoOtelAdapter`

packages/shared-api/middlewares/logging.ts
```diff
@@ -0,0 +1,36 @@
...
+export const logger = (
+  serviceName: string,
+  logger = createPinoAdapter(getOtelLogger(serviceName, logLevel)),
+) => {
+  const app = new Elysia({ name: "hebo-logging" }).decorate("logger", logger);
+
+  if (!isProduction) {
+    app.onRequest(({ request }) => {
+      logger.info(
+        { method: request.method, path: getPathnameFromUrl(request.url) },
```
heiwen: `getPathnameFromUrl` is not necessary anymore. Since this code only executes in development, you can just use `new Url()`.

packages/shared-api/lib/otel.ts
heiwen: If it is misspelled, this would swallow the mistake and make the user very confused. Minimum would be to warn. In this case wouldn't even start the app, since it's an obvious error.

turisanapo: Added an explicit error to prevent users from running the application with misspelt log levels.

packages/shared-api/utils/otel/log-levels.ts
turisanapo: Good point, updated.

packages/shared-api/middlewares/logging.ts
turisanapo: Removed

apps/gateway/src/gateway-config.ts
turisanapo: Renamed

PR #262, opened 2026-03-18
fix: use React Router Link in pagination for client-side navigation

### LINKED ISSUE #260: Bug: Traces pagination causes full page refresh instead of client-side navigation
**Area:** `packages/shared-ui/src/_shadcn/ui/pagination.tsx`, `apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.traces/list.tsx`

## Background

Clicking "Next" or "Previous" in the traces list triggers a full browser page reload. The pagination controls in `list.tsx` pass an `href` prop to `PaginationNext` / `PaginationPrevious`, which ultimately renders a native `<a>` tag ([pagination.tsx:54](../packages/shared-ui/src/_shadcn/ui/pagination.tsx#L54)). Native anchors bypass React Router, so the app re-mounts from scratch on every page turn instead of doing a client-side navigation.

## Root cause

`PaginationLink` uses `render={<a ... />}` inside the shadcn `Button`. React Router's `<Link>` component is never used, so the router never intercepts the click.

## Proposed fix

Replace the native `<a>` render prop in `PaginationLink` with React Router's `<Link>` component:

```tsx
import { Link } from "react-router";

// in PaginationLink
render={
  <Link
    aria-current={isActive ? "page" : undefined}
    data-slot="pagination-link"
    data-active={isActive}
    {...props}
  />
}
```

`Link` accepts the same `href`-equivalent (`to`) and all standard anchor props, so no changes are needed at call sites in `list.tsx` — provided `Link` accepts an `href` prop as an alias for `to`, or call sites are updated to pass `to` instead.

> Check whether React Router v7's `<Link>` accepts `href` directly. If not, add a thin wrapper that maps `href` → `to`.

## Acceptance criteria

- [ ] Follow existing pattern to extend components in our own wrapper instead of changing shadcn files
- [ ] Clicking "Next" / "Previous" in the traces list does not cause a full page reload
- [ ] URL search params (`page`, `from`, `to`, etc.) update correctly after navigation
- [ ] Previously selected trace and scroll position are not lost on pagination (soft navigation)
- [ ] "Previous" and "Next" work correctly at the first and last pages (disabled state unchanged)

heiwen:
Replace native `<a>` with React Router `<Link>` in PaginationLink wrapper, mapping `href` → `to`. Stays close to the original shadcn component structure with no call-site changes.

Closes #260

## Summary by CodeRabbit

* **New Features**
  * Pagination components implemented locally with improved active styling, ARIA attributes, and adaptive text/icon rendering.

* **Bug Fixes**
  * Paginated views now remount correctly when switching pages to avoid stale content.

* **Chores**
  * Added routing and auth-related packages to the top-level workspace catalog and updated package declarations to reference catalog entries; added react-router to shared UI peers.

* **Documentation**
  * Added guidance for promoting cross-package dependencies to workspace-level catalog entries.


heiwen:
@claude once we use a dependency in more than one app / package, promote it to a workspace dependency. that includes all it's related dependencies that follow the same versioning. Do that and Add this as a rule to agents.md as well.

packages/shared-ui/src/components/Pagination.tsx
```diff
@@ -1,9 +1,81 @@
...
+      className={cn(className)}
+      nativeButton={false}
+      render={
+        <Link
+          aria-current={isActive ? "page" : undefined}
+          data-slot="pagination-link"
+          data-active={isActive}
+          to={href ?? "#"}
+          {...props}
+        />
```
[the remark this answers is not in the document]
heiwen: Same as the original component, we are fine with that.

heiwen:
@claude please address codex concern

heiwen:
@turisanapo ready for final review

packages/shared-ui/package.json
```diff
@@ -32,6 +32,7 @@
     "lucide-react": "catalog:",
     "react": "catalog:",
     "react-dom": "catalog:",
+    "react-router": "catalog:",
```
turisanapo, line 35: `shared-ui` could be consumed outside of the console. Is adding `react-router` as a hard dependency a limitation we should address?

packages/shared-ui/src/components/Pagination.tsx
```diff
@@ -1,9 +1,81 @@
...
+  isActive,
+  size = "icon",
+  href,
+  ...props
+}: PaginationLinkProps) {
+  return (
+    <Button
+      variant={isActive ? "outline" : "ghost"}
+      size={size}
+      className={cn(className)}
```
turisanapo: You should be able to pass `className` directly here.

packages/shared-ui/package.json
heiwen: Shared-ui will only be consumed by our own frontends, all of them using react router.

heiwen:
@claude please look at Davide's last comment

turisanapo (approved):

PR #297, opened 2026-03-19
fix: no exception on optional columns

### LINKED ISSUE #296: PostgresError: Failed to plan SQL: No field named \"span_attributes.gen_ai.usage.cache_read.input_tokens\".
The issue is that GreptimeDB creates columns dynamically when data is first inserted. If no span with gen_ai.usage.cache_read.input_tokens has been received yet, that column doesn't exist in the schema and the query fails at plan time.

heiwen:
Fixes: https://github.com/8monkey-ai/hebo-platform/issues/296

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metadata column detection to handle missing or differently-named columns gracefully.
  * Optional token fields are now included conditionally, preventing query errors when those columns are absent.
  * Trace and span retrieval is more robust across varying database schemas, yielding consistent results even if optional columns are unavailable.

### DIFF
--- a/apps/api/src/modules/traces/service.ts
+++ b/apps/api/src/modules/traces/service.ts
@@ -35,7 +35,11 @@ async function getMetadataColumnNames(greptimeDb: GreptimeDb) {
     `SELECT column_name
      FROM information_schema.columns
      WHERE table_name = 'opentelemetry_traces'
-       AND column_name LIKE '${METADATA_PREFIX}%'`,
+       AND (column_name LIKE '${METADATA_PREFIX}%'
+         OR column_name IN (
+           'span_attributes.gen_ai.usage.cache_read.input_tokens',
+           'span_attributes.gen_ai.usage.reasoning.output_tokens'
+         ))`,
   )) as Array<{ column_name: unknown }>;
   metadataColumnsCache.set("metadata_columns", result);
   return result;
@@ -54,7 +58,9 @@ export async function listTraces(
   statusFilter?: "ok" | "error",
   operationFilter?: "chat" | "embeddings",
 ) {
-  const metadataColumns = await getMetadataColumnNames(greptimeDb);
+  const metadataColumns = (await getMetadataColumnNames(greptimeDb)).filter(({ column_name }) =>
+    String(column_name).startsWith(METADATA_PREFIX),
+  );
   const metadataKeys = metadataColumns.map(({ column_name }) =>
     String(column_name).slice(METADATA_PREFIX.length),
   );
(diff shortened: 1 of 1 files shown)


apps/api/src/modules/traces/service.ts
turisanapo: `getMetadataColumnNames` is doing 2 things now.

It returns both metadata columns and the two optional token columns, and callers have to `.filter()`.

I suggest renaming it to something like `getTraceColumnNames` for clarity and make it return a structured object (e.g. `{ metadataColumns, optionalColumns }`) so callers don't need to re-filter.

apps/api/src/modules/traces/service.ts
```diff
@@ -35,7 +35,11 @@ async function getMetadataColumnNames(greptimeDb: GreptimeDb) {
     `SELECT column_name
      FROM information_schema.columns
      WHERE table_name = 'opentelemetry_traces'
-       AND column_name LIKE '${METADATA_PREFIX}%'`,
+       AND (column_name LIKE '${METADATA_PREFIX}%'
+         OR column_name IN (
+           'span_attributes.gen_ai.usage.cache_read.input_tokens',
+           'span_attributes.gen_ai.usage.reasoning.output_tokens'
```
turisanapo: These 2 special columns are hardcoded in both here and the `optionalCol` calls. Adding a third would mean updating two spots.

What about extracting a constant like `OPTIONAL_SPAN_COLUMNS = [{ column, alias }]` and deriving both from it?

apps/api/src/modules/traces/service.ts
```diff
@@ -35,7 +35,11 @@ async function getMetadataColumnNames(greptimeDb: GreptimeDb) {
...
+         ))`,
   )) as Array<{ column_name: unknown }>;
   metadataColumnsCache.set("metadata_columns", result);
```
turisanapo: Nit: it is not `"metadata_columns"` only anymore. Suggest renaming it to `"trace_columns"`

heiwen:
@claude what do you think about davide's comments?

heiwen:
@claude do it

turisanapo:
@claude

1/ The `TraceColumns` interface and the explicit :` Promise<TraceColumns>` return annotation aren't necessary. TypeScript infers the return shape from the object literal, and the cache picks it up too. Suggest to drop both to keep things simple.

2/ `const optionalColumnsIn` is only used once. Can we inline it?

3/ `const result: TraceColumns` can be simplified: the `.map(({ column, alias }) => ({ column, alias }))` on `OPTIONAL_SPAN_COLUMNS` just reconstructs the same shape and the filter result can be used directly. And for `metadataColumns`, the double `String(column_name)` in filter + map could be a single-pass reduce or flatMap.

turisanapo (approved):

PR #302, opened 2026-03-20
fix: add invited members to agent's team using stored team_id

### LINKED ISSUE #301: Bug: Adding a member to an agent does not add them to the agent's team
## Bug

When a member is invited to an agent, they are added to the organization but not to the agent's team.

## Root cause

`authService.inviteMember` calls `organization.inviteMember` (adds to org only). It does not call `organization.addTeamMember`, so the new member is never added to the team that was created for the agent.

Relevant code:
- `apps/console/app/lib/auth/better-auth.ts` — `inviteMember()` (line ~206)
- `apps/api/src/modules/agents.ts` — agent creation correctly adds the creator to the team via `addTeamMember` (line ~53)

## Expected behavior

When a member is added to an agent, they should also be added to that agent's team (identified by the agent's `team_id`).

## Steps to reproduce

1. Create an agent (creator is added to the team)
2. Invite another user to the agent
3. Check team membership — the invited user is not in the agent's team

heiwen:
Pass the agent's `team_id` through the invite form as a hidden field instead of re-querying the agent on submit. The `team_id` is already available from the parent route's loader data.

Closes #301

## Summary by CodeRabbit

* **New Features**
  * Team-scoped member invitations: invites now include team context, are submitted with team info, and pending invites are shown per team.
  * Invite form and members list automatically associate and filter invitations by the current agent/team.

* **Bug Fixes**
  * Improved handling of pending invitations to display consistent fields and support resend behavior.

### DIFF
--- a/apps/api/prisma/schema.prisma
+++ b/apps/api/prisma/schema.prisma
@@ -44,7 +44,7 @@ model agents {
   deleted_at      DateTime?
   /// @prismabox.hide
   organization_id String
-  /// @prismabox.hide
+  /// @prismabox.input.hide
   team_id         String
 }
 
--- a/apps/console/app/lib/auth/better-auth.ts
+++ b/apps/console/app/lib/auth/better-auth.ts
@@ -188,16 +188,9 @@ export const authService: AuthService = {
       }
     }
 
-    const invitations = ((data.invitations ?? []) as unknown as (OrgInvitation & Record<string, unknown>)[])
-      .filter((i) => i.status === "pending")
-      .map(({ id, email, role, expiresAt, status, teamId }) => ({
-        id,
-        email,
-        role,
-        expiresAt,
-        status,
-        teamId: typeof teamId === "string" ? teamId : undefined,
-      }));
+    const invitations = ((data.invitations ?? []) as unknown as OrgInvitation[]).filter(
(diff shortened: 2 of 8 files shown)


heiwen:
@turisanapo for review

heiwen:
@claude what do you think about the latest comments?

heiwen:
@claude for (1), how do we handle that in other forms? I think (2) is worth fixing. let's be a good citizen and add resend for (3).

heiwen:
@claude you are scoping to the teamId in the frontend. Are you sure that is the right / secure way to do things?

heiwen:
@claude got it, we're fine then from a security perspective.

But I guess this additional map that you introduced here is unnessecary

```
    const invitations = ((data.invitations ?? []) as unknown as (OrgInvitation & Record<string, unknown>)[])
      .filter((i) => i.status === "pending")
      .map(({ id, email, role, expiresAt, status, teamId }) => ({
        id,
        email,
        role,
        expiresAt,
        status,
        teamId: typeof teamId === "string" ? teamId : undefined,
      }));
```

and the filtering itself should already happen in the client loader of the agent settings page?

heiwen:
@turisanapo for final review

turisanapo (approved):

PR #306, opened 2026-03-20
feat: add Bedrock access key, Vertex service account, and Azure OpenAI support

### LINKED ISSUE #300: Gateway provider expansion: Bedrock static creds, Vertex service account, Azure OpenAI
## Context

Prerequisite for #298 (self-hosted Docker distribution). Self-hosted users won't have ECS container metadata or AWS→GCP Workload Identity Federation. They need simple credential-based auth paths for cloud providers.

## Background

The gateway (`apps/gateway`) currently supports 6 providers, but Bedrock and Vertex use cloud-native auth that won't work in self-hosted:

| Provider | Current auth method | Self-hosted ready? |
|---|---|---|
| **Anthropic** | API key | Yes |
| **OpenAI** | API key | Yes |
| **Groq** | API key | Yes |
| **Voyage** | API key | Yes |
| **Bedrock** | IAM Role ARN + ECS `fromContainerMetadata()` → STS `AssumeRole` | **No** |
| **Vertex AI** | AWS→GCP Workload Identity Federation | **No** |
| **Azure OpenAI** | *(not supported)* | N/A |

### 1. Bedrock — add static credentials support

`apps/gateway/src/services/provider-factory.ts:81-98` currently only supports `fromTemporaryCredentials` + `fromContainerMetadata`. Add an alternative path using `accessKeyId` + `secretAccessKey` + `region`.

The `@ai-sdk/amazon-bedrock` SDK already supports static credentials. Add a new config shape:

```typescript
// In apps/api/src/modules/providers/types.ts
const BedrockStaticProviderConfig = t.Object({
  accessKeyId: t.String({ "x-redact": true }),
  secretAccessKey: t.String({ "x-redact": true }),
  region: t.String(),
});
```

In `createProvider()`, if `accessKeyId` + `secretAccessKey` are provided, use static credentials instead of the IAM role path. The existing IAM role path remains for cloud deployments.

### 2. Vertex AI — add service account JSON key support

`apps/gateway/src/services/provider-factory.ts:105-117` currently only supports WIF via `buildWifOptions()`. Add support for GCP service account JSON credentials.

```typescript
// In apps/api/src/modules/providers/types.ts
const VertexServiceAccountProviderConfig = t.Object({
  serviceAccountKey: t.String({ "x-redact": true }),  // JSON key as string
  location: t.String(),
  project: t.String(),
});
```

The `@ai-sdk/google-vertex` SDK supports `googleAuthOptions.credentials` with parsed JSON credentials directly.

### 3. Google AI (Gemini via API key) — consider as simpler alternative

Consider adding `@ai-sdk/google` as a new provider for users who want Gemini models without Vertex AI setup. This is a simple API key provider like Anthropic/OpenAI.

### 4. Azure OpenAI — add as new provider

Add `azure` as a new provider using `@ai-sdk/azure`:

```typescript
// In apps/api/src/modules/providers/types.ts
const AzureProviderConfig = t.Object({
  apiKey: t.String({ "x-redact": true }),
  resourceName: t.String(),
  apiVersion: t.Optional(t.String()),
});
```

Requires:
- Adding `@ai-sdk/azure` dependency
- New `azure` entry in `supportedProviders` (`apps/api/src/modules/providers/types.ts`)
- New case in `createProvider()` (`apps/gateway/src/services/provider-factory.ts`)
- Console UI form fields in `configure.tsx`

### 5. Enable more default models

Current model catalog (`apps/gateway/src/gateway-config.ts:49-78`) has models from OpenAI, Anthropic, Google, Amazon, and Voyage. Consider expanding defaults to include models accessible via the new provider paths (e.g., Azure OpenAI models, Gemini via Google AI).

## Files affected

- `apps/api/src/modules/providers/types.ts` — add new config schemas (`BedrockStaticProviderConfig`, `VertexServiceAccountProviderConfig`, `AzureProviderConfig`), update `ProviderConfig` union, add `azure` to `supportedProviders`
- `apps/gateway/src/services/provider-factory.ts` — add static credential paths for Bedrock, Vertex; add Azure case
- `apps/gateway/src/gateway-config.ts` — add Azure provider, potentially add Google AI provider
- `apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.providers/configure.tsx` — add form fields for new config shapes (Bedrock static creds, Vertex service account, Azure)
- `apps/console/app/mocks/routes/providers.ts` — update mocks for new providers
- `package.json` — add `@ai-sdk/azure` (and potentially `@ai-sdk/google`) dependencies

## Acceptance criteria

- [ ] Bedrock works with static `accessKeyId` + `secretAccessKey` + `region` (existing IAM role path still works)
- [ ] Vertex works with service account JSON key (existing WIF path still works)
- [ ] Azure OpenAI provider works with `apiKey` + `resourceName`
- [ ] Console BYOK form renders correct fields for each provider config shape
- [ ] All existing providers continue to work unchanged
- [ ] Tests cover new provider creation paths

## Related

- Parent issue: #298
- Ticket 1: Password-based auth flow (#299)

heiwen:
Adds `authMode` discriminated union for Bedrock (iam-role | static) and Vertex (wif | service-account) to support self-hosted deployments. Adds Azure OpenAI as a new provider. Console UI uses tabs to switch auth modes.

Closes #300

## Summary by CodeRabbit

* **New Features**
  * Added Azure provider (resource API-key + resource selection), multiple auth modes for Bedrock and Vertex, and auth-mode tabs in provider UI; provider icons now show Microsoft for Azure
  * Added a reusable Textarea component for multiline inputs

* **Tests**
  * Added provider-factory tests covering new providers and auth modes

* **Style**
  * Improved label formatting for hyphenated keys

* **Chores**
  * Gateway configuration updated to register Azure and adopt authMode-aware provider definitions

### DIFF
--- a/apps/api/src/modules/providers/types.ts
+++ b/apps/api/src/modules/providers/types.ts
@@ -54,7 +54,7 @@ const ApiKeyProviderConfig = t.Object({
 });
 
 const AzureProviderConfig = t.Object({
-  authMode: t.Literal("api-key"),
+  authMode: t.Literal("azure-api-key"),
   apiKey: t.String({ "x-redact": true }),
   resourceName: t.String(),
   apiVersion: t.Optional(t.String()),
--- a/apps/api/src/modules/providers/types.ts
+++ b/apps/api/src/modules/providers/types.ts
@@ -48,10 +48,12 @@ const VertexServiceAccountConfig = t.Object({
 const VertexProviderConfig = t.Union([VertexIdentityFederationConfig, VertexServiceAccountConfig]);
 
 const ApiKeyProviderConfig = t.Object({
+  authMode: t.Literal("api-key"),
   apiKey: t.String({ "x-redact": true }),
 });
 
 const AzureProviderConfig = t.Object({
+  authMode: t.Literal("api-key"),
   apiKey: t.String({ "x-redact": true }),
   resourceName: t.String(),
   apiVersion: t.Optional(t.String()),
--- a/apps/api/src/modules/providers/types.ts
+++ b/apps/api/src/modules/providers/types.ts
@@ -40,7 +40,8 @@ const VertexIdentityFederationConfig = t.Object({
 
 const VertexServiceAccountConfig = t.Object({
   authMode: t.Literal("service-account"),
-  serviceAccountKey: t.String({ "x-redact": true }),
+  clientEmail: t.String(),
+  privateKey: t.String({ "x-redact": true }),
   location: t.String(),
   project: t.String(),
 });
--- a/apps/api/src/modules/providers/types.ts
+++ b/apps/api/src/modules/providers/types.ts
@@ -21,17 +21,17 @@ const BedrockIamRoleConfig = t.Object({
   region: t.String(),
 });
 
-const BedrockStaticConfig = t.Object({
-  authMode: t.Literal("static"),
+const BedrockAccessKeyConfig = t.Object({
+  authMode: t.Literal("access-key"),
   accessKeyId: t.String({ "x-redact": true }),
   secretAccessKey: t.String({ "x-redact": true }),
   region: t.String(),
 });
 
-const BedrockProviderConfig = t.Union([BedrockIamRoleConfig, BedrockStaticConfig]);
+const BedrockProviderConfig = t.Union([BedrockIamRoleConfig, BedrockAccessKeyConfig]);
 
-const VertexWifConfig = t.Object({
-  authMode: t.Literal("wif"),
+const VertexIdentityFederationConfig = t.Object({
+  authMode: t.Literal("identity-federation"),
   serviceAccountEmail: t.String(),
   audience: t.String(),
   location: t.String(),
@@ -45,7 +45,7 @@ const VertexServiceAccountConfig = t.Object({
   project: t.String(),
 });
 
-const VertexProviderConfig = t.Union([VertexWifConfig, VertexServiceAccountConfig]);
+const VertexProviderConfig = t.Union([VertexIdentityFederationConfig, VertexServiceAccountConfig]);
 
 const ApiKeyProviderConfig = t.Object({
   apiKey: t.String({ "x-redact": true }),
@@ -82,9 +82,9 @@ export const Models = t.Array(
 
 export type Models = Static<typeof Models>;
 export type BedrockIamRoleConfig = Static<typeof BedrockIamRoleConfig>;
-export type BedrockStaticConfig = Static<typeof BedrockStaticConfig>;
+export type BedrockAccessKeyConfig = Static<typeof BedrockAccessKeyConfig>;
 export type BedrockProviderConfig = Static<typeof BedrockProviderConfig>;
-export type VertexWifConfig = Static<typeof VertexWifConfig>;
+export type VertexIdentityFederationConfig = Static<typeof VertexIdentityFederationConfig>;
 export type VertexServiceAccountConfig = Static<typeof VertexServiceAccountConfig>;
 export type VertexProviderConfig = Static<typeof VertexProviderConfig>;
 export type ApiKeyProviderConfig = Static<typeof ApiKeyProviderConfig>;
--- a/apps/api/src/modules/providers/types.ts
+++ b/apps/api/src/modules/providers/types.ts
@@ -57,7 +57,6 @@ const AzureProviderConfig = t.Object({
   authMode: t.Literal("azure-api-key"),
   apiKey: t.String({ "x-redact": true }),
   resourceName: t.String(),
-  apiVersion: t.Optional(t.String()),
 });
 
 export const ProviderConfig = t.Union([
--- a/apps/api/src/modules/providers/types.ts
+++ b/apps/api/src/modules/providers/types.ts
@@ -2,6 +2,7 @@ import { t, type Static } from "elysia";
 
 export const supportedProviders = {
   anthropic: { name: "Anthropic" },
+  azure: { name: "Azure OpenAI" },
   bedrock: { name: "Amazon Bedrock" },
   groq: { name: "Groq" },
   openai: { name: "OpenAI" },
@@ -14,26 +15,53 @@ export const ProviderSlug = t.Enum(
   { error: "Invalid provider slug" },
 );
 
-const BedrockProviderConfig = t.Object({
+const BedrockIamRoleConfig = t.Object({
+  authMode: t.Literal("iam-role"),
   bedrockRoleArn: t.String(),
   region: t.String(),
 });
 
-const VertexProviderConfig = t.Object({
+const BedrockStaticConfig = t.Object({
+  authMode: t.Literal("static"),
+  accessKeyId: t.String({ "x-redact": true }),
+  secretAccessKey: t.String({ "x-redact": true }),
+  region: t.String(),
+});
+
+const BedrockProviderConfig = t.Union([BedrockIamRoleConfig, BedrockStaticConfig]);
+
+const VertexWifConfig = t.Object({
+  authMode: t.Literal("wif"),
   serviceAccountEmail: t.String(),
   audience: t.String(),
   location: t.String(),
   project: t.String(),
 });
 
+const VertexServiceAccountConfig = t.Object({
+  authMode: t.Literal("service-account"),
+  serviceAccountKey: t.String({ "x-redact": true }),
+  location: t.String(),
+  project: t.String(),
+});
+
+const VertexProviderConfig = t.Union([VertexWifConfig, VertexServiceAccountConfig]);
+
 const ApiKeyProviderConfig = t.Object({
   apiKey: t.String({ "x-redact": true }),
+});
+
+const AzureProviderConfig = t.Object({
+  apiKey: t.String({ "x-redact": true }),
+  resourceName: t.String(),
+  apiVersion: t.Optional(t.String()),
 });
 
 export const ProviderConfig = t.Union([
   BedrockProviderConfig,
   VertexProviderConfig,
   ApiKeyProviderConfig,
+  AzureProviderConfig,
 ]);
 
 export const Provider = t.Object({
@@ -53,9 +81,14 @@ export const Models = t.Array(
 );
 
 export type Models = Static<typeof Models>;
+export type BedrockIamRoleConfig = Static<typeof BedrockIamRoleConfig>;
+export type BedrockStaticConfig = Static<typeof BedrockStaticConfig>;
 export type BedrockProviderConfig = Static<typeof BedrockProviderConfig>;
+export type VertexWifConfig = Static<typeof VertexWifConfig>;
+export type VertexServiceAccountConfig = Static<typeof VertexServiceAccountConfig>;
 export type VertexProviderConfig = Static<typeof VertexProviderConfig>;
 export type ApiKeyProviderConfig = Static<typeof ApiKeyProviderConfig>;
+export type AzureProviderConfig = Static<typeof AzureProviderConfig>;
 export type Provider = Static<typeof Provider>;
 export type ProviderConfig = Static<typeof ProviderConfig>;
 export type ProviderSlug = Static<typeof ProviderSlug>;
--- a/apps/api/src/modules/providers/types.ts
+++ b/apps/api/src/modules/providers/types.ts
@@ -54,9 +54,9 @@ const ApiKeyProviderConfig = t.Object({
 });
 
 const AzureProviderConfig = t.Object({
-  authMode: t.Literal("azure-api-key"),
-  apiKey: t.String({ "x-redact": true }),
+  authMode: t.Literal("resource-api-key"),
   resourceName: t.String(),
+  apiKey: t.String({ "x-redact": true }),
 });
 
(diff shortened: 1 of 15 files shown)


apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.providers/configure.tsx
```diff
@@ -62,11 +89,54 @@ type ConfigureProviderDialogProps = {
...
+      return authMode === "service-account"
+        ? ["serviceAccountKey", "location", "project"]
+        : ["serviceAccountEmail", "audience", "location", "project"];
+    case "azure":
+      return ["apiKey", "resourceName", "apiVersion"];
+    default:
+      return ["apiKey"];
+  }
+}
+
```
heiwen: Can we simplify this part, e.g. derive it from the zod schema right away? And / or have authModes for all providers, no matter whether they have multiple or a single, and then just decide no tabs based on only a single auth mode being present?

heiwen:
@claude have a look at my PR comment

heiwen:
@claude I believe this can be further simplifed by (i) using the existing zod schema instead of a new type (you might infer the type from that if needed) (ii) not having a default, but being explicit about each provider

heiwen:
@claude I'm still wondering whether we need `providerAuthModes` at all, or can just work with `ProviderConfigureSchema` (or a type derived from that)?

heiwen:
@claude no label map, you can UI-ify that (i.e. split by "-", capitalise, ...). Derive everything from schema.

heiwen:
@claude let's rename "wif" to "identify-federation" and "static" to "access-key" (also in the types)

heiwen:
@claude shouldn't `azure` and the `apikey` modes also have a `authMode` type so that the discriminatedUnion is correct?

heiwen:
@claude any backwards compatibility concerns with this approach?

apps/api/src/modules/providers/types.ts
```diff
@@ -14,26 +15,55 @@ export const ProviderSlug = t.Enum(
...
+  serviceAccountKey: t.String({ "x-redact": true }),
+  location: t.String(),
+  project: t.String(),
+});
+
+const VertexProviderConfig = t.Union([VertexIdentityFederationConfig, VertexServiceAccountConfig]);
+
 const ApiKeyProviderConfig = t.Object({
+  authMode: t.Literal("api-key"),
+  apiKey: t.String({ "x-redact": true }),
```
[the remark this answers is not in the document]
heiwen: We'll do a separate migration

heiwen:
@claude what do you think about concern (1)? for concern (2), I would actually like to avoid a json value in a text field. what are the minimum fields from that json that we need? we should make them their own fields in the schema (and with that text fields). Is the actual key actually enough? any other areas where we still accept json?

apps/api/src/modules/providers/types.ts
```diff
@@ -14,26 +15,55 @@ export const ProviderSlug = t.Enum(
...
+});
+
+const AzureProviderConfig = t.Object({
+  authMode: t.Literal("api-key"),
   apiKey: t.String({ "x-redact": true }),
+  resourceName: t.String(),
+  apiVersion: t.Optional(t.String()),
 });
```
heiwen, line 63: Will do separately

heiwen:
@claude please implement both fixes (the empty modes guard + replacing serviceAccountKey with clientEmail/privateKey)

heiwen:
@claude I think there is one more new comment from CodeRabbit. What do you think?

heiwen:
@claude give azure its own type then

heiwen:
@claude two more comments, thoughts?

heiwen:
@claude 
1. I guess the decision input vs text area should be based on the schema, how would you do it?
2. Do we really need the aversion?

heiwen:
@claude
For 1/ I think the cleaner way is to use zod .extension(), since describe is meant for visible text
For 2/ lets remove the APIversion then

heiwen:
@claude

I meant to say use "meta" from zod for the textarea extension: https://zod.dev/metadata

What about the two new comments from CodeRabbit?

heiwen:
@claude no need for global registry

heiwen:
@claude look at the latest two comments from coderabbit

apps/gateway/src/services/provider-factory.ts
```diff
@@ -79,8 +87,22 @@ export async function loadProviderSecrets() {
...
     case "bedrock": {
-      const { bedrockRoleArn, region } = config as BedrockProviderConfig;
-      if (!bedrockRoleArn || !region) return;
+      const bedrockConfig = config as BedrockProviderConfig;
+      const region = bedrockConfig.region;
+      if (!region) return;
+
+      if ("authMode" in bedrockConfig && bedrockConfig.authMode === "access-key") {
+        const { accessKeyId, secretAccessKey } = bedrockConfig;
+        if (!accessKeyId || !secretAccessKey) return;
```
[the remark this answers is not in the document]
heiwen: @coderabbit is this still an issue?

apps/gateway/src/services/provider-factory.ts
```diff
@@ -103,8 +125,31 @@ export function createProvider(slug: ProviderSlug, config: unknown): ProviderV3
...
+        return withCanonicalIdsForVertex(
+          createVertex({
+            googleAuthOptions: {
+              credentials,
+              scopes: ["https://www.googleapis.com/auth/cloud-platform"],
+            },
+            location,
+            project,
+          }),
+        );
```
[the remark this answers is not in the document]
heiwen: We don't use JSON anymore, closing

heiwen:
@claude summarize the compatibility / migration concerns again

heiwen:
@claude did we forget to add the azure icon?

apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.providers/configure.tsx
```diff
@@ -82,16 +155,50 @@ export function ConfigureProviderDialog({ provider, ...props }: ConfigureProvide
...
+      {(activeKeys as (keyof typeof configFieldset)[]).map((key) => {
+        const field = configFieldset[key];
+        if (!field) return null;
+        return (
+          <Field key={key} name={field.name}>
+            <FieldLabel>{labelize(key)}</FieldLabel>
+            <FieldControl>
+              {activeMode && isTextarea(activeMode.schema, key) ? (
+                <textarea
```
turisanapo: Have you considered using a ShadCN Textarea component instead? It would remove the need for the inline class string

apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.providers/configure.tsx
```diff
@@ -23,36 +23,71 @@ import {
...
+  bedrockRoleArn: requiredString("Please enter a valid Bedrock ARN role"),
+  region: requiredString("Please enter a valid AWS region"),
+});
+
+const BedrockAccessKeySchema = z.object({
+  authMode: z.literal("access-key"),
+  accessKeyId: requiredString("Please enter a valid access key ID"),
+  secretAccessKey: requiredString("Please enter a valid secret access key"),
+  region: requiredString("Please enter a valid AWS region"),
+});
```
turisanapo, line 47: Have you considered using a single source of truth for the provider config schemas defined here and in apps/api/src/modules/providers/types.ts? 

How can we prevent schemas from drifting over time?

apps/gateway/src/services/provider-factory.ts
```diff
@@ -131,6 +177,11 @@ export function createProvider(slug: ProviderSlug, config: unknown): ProviderV3
       if (!apiKey) return;
       return withCanonicalIdsForOpenAI(createOpenAI({ apiKey }));
     }
+    case "azure": {
+      const { apiKey, resourceName } = config as AzureProviderConfig;
+      if (!apiKey || !resourceName) return;
+      return withCanonicalIdsForOpenAI(createAzure({ apiKey, resourceName }));
```
turisanapo: Is using `withCanonicalIdsForOpenAI` for the Azure provider intentional? Model IDs in Azure do not match the OpenAI ones.

apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.providers/configure.tsx
heiwen: We talked about this before. The issue is that we use different schema libraries in frontend and backend. This is a historic artifact, from when Elysia didn't support standard schema yet, which it supports now. The way would be to unify that, but that's out of scope of this PR, it would impact almost all routes.

apps/gateway/src/services/provider-factory.ts
heiwen: That's a mistake yes. Azure doesn't need any canonical wrapper. Instead, the models need to be configured on Azure itself to match the canonical names.

apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.providers/configure.tsx
heiwen: Good idea

heiwen:
@claude please address @turisanapo's comments

apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.providers/configure.tsx
turisanapo: Opened an issue to keep track: https://github.com/8monkey-ai/hebo-platform/issues/308

heiwen:
@coderabbitai review the latest set of changes

heiwen:
@claude what do you think about the latest code rabbit comments

heiwen:
@claude would reading the private key be the right place for the normalization though, or should it be normalized either in the form itself or in the api on write?

heiwen:
@claude let's use the typebox transform then and also fix the test. We'll leave 1, 3 & 4 as is.

heiwen:
@claude if we add minLength: 1 to credential string fields, we can get rid of the falsy checks in the factory?

apps/api/src/modules/providers/types.ts
```diff
@@ -14,26 +15,55 @@ export const ProviderSlug = t.Enum(
...
 const ApiKeyProviderConfig = t.Object({
+  authMode: t.Literal("api-key"),
+  apiKey: t.String({ "x-redact": true }),
+});
+
+const AzureProviderConfig = t.Object({
+  authMode: t.Literal("resource-api-key"),
+  resourceName: t.String(),
   apiKey: t.String({ "x-redact": true }),
 });
```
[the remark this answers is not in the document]
heiwen: We'll skip for now.

turisanapo:
## Migration: backfill `authMode` on existing provider configs

This PR makes `authMode` a required field on all provider config schemas. Existing `provider_configs` rows in the database were written without it and will fail schema validation on read.

### Recommended approach

Run the following SQL **once per environment** before or immediately after deploying this version. The gateway handles missing `authMode` gracefully during the transition (falls back to the original auth path), so there is no downtime risk if the deploy happens slightly before the script.

```sql
UPDATE api.provider_configs
SET value = value || CASE
    WHEN provider_slug = 'bedrock' THEN '{"authMode": "iam-role"}'::jsonb
    WHEN provider_slug = 'vertex' THEN '{"authMode": "identity-federation"}'::jsonb
    ELSE '{"authMode": "api-key"}'::jsonb
  END
WHERE NOT (value ? 'authMode')
  AND deleted_at IS NULL;
```

**Notes:**
- Idempotent — safe to run multiple times (`NOT (value ? 'authMode')` skips already-migrated rows)
- Azure is excluded because it's a new provider introduced by this PR (no pre-existing rows)
- The `ELSE` branch covers all api-key providers (anthropic, openai, groq, voyage)
- Validated locally against existing dev data (bedrock, groq, voyage rows)

This should be included in the release notes for the version that ships this PR.

turisanapo:
@claude, what do you think about the proposed migration approach?

apps/gateway/src/gateway-config.ts
```diff
@@ -30,20 +30,30 @@ export const gw = gateway({
...
-    voyage: createProvider("voyage", { apiKey: secrets.voyageApiKey }),
-    anthropic: createProvider("anthropic", { apiKey: secrets.anthropicApiKey }),
-    openai: createProvider("openai", { apiKey: secrets.openAiApiKey }),
+    voyage: createProvider("voyage", { authMode: "api-key", apiKey: secrets.voyageApiKey }),
+    anthropic: createProvider("anthropic", {
+      authMode: "api-key",
+      apiKey: secrets.anthropicApiKey,
+    }),
+    openai: createProvider("openai", { authMode: "api-key", apiKey: secrets.openAiApiKey }),
+    azure: createProvider("azure", {
```
turisanapo, line 52: Is there a reason why we are defining the `"azure"` provider but not linking it to any model? Anthropic models are available on Azure as well.

heiwen:
@coderabbit review the latest set of changes

turisanapo (approved):

PR #319, opened 2026-03-29
chore: add typecheck step to pre-commit and CI workflow

turisanapo:
## Summary

Add `bun run typecheck` (`oxlint --type-aware --type-check`) to pre-commit hooks and CI quality workflow so type errors are caught before merge.

## Changes

- **Pre-commit hook** — added a `typecheck` command to `lefthook.yml` that runs alongside format and lint on staged `.ts/.tsx` files.
- **CI workflow** — added "Run typecheck" and "Run tests" steps to `quality.yml`.
- **Parallel typecheck with `pretypecheck`** — added a `pretypecheck` lifecycle hook to the root `package.json` that runs `bun run predev` before typecheck. This generates Prisma clients and route types upfront so all workspace typechecks can run in parallel — without it, workspaces like `@hebo/gateway` fail because they import types from `@hebo/api` that depend on generated Prisma types. Each workspace also keeps `bun run predev` in its own `typecheck` script so single-workspace runs (`bun run -F @hebo/api typecheck`) still work standalone.
- **`@sinclair/typebox` as dev dependency** — added to `apps/api`. It's imported as `import type` in `redact-provider.ts` for type-safe schema traversal. Without it, the type-aware linter can't resolve the module after a clean install since bun's peer dep hoisting from `elysia` is not guaranteed.

## Summary by CodeRabbit

* **Chores**
  * Enhanced continuous integration pipeline with automated type checking and test execution for improved code quality.
  * Integrated type checking into pre-commit hooks to catch issues earlier in the development workflow.
  * Added development dependencies to support enhanced tooling capabilities.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -31,6 +31,7 @@
   },
   "devDependencies": {
     "@hebo/typescript-config": "workspace:*",
+    "@sinclair/typebox": "^0.34.49",
     "@types/bun": "catalog:",
     "prisma": "catalog:",
     "prismabox": "^1.1.26"
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -16,7 +16,7 @@
     "format": "oxfmt .",
     "lint": "oxlint",
     "test": "bun test",
-    "typecheck": "oxlint --type-aware --type-check",
+    "typecheck": "bun run predev && oxlint --type-aware --type-check",
     "prestart": "bun run build",
     "start": "./dist/server"
   },
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -16,7 +16,7 @@
     "format": "oxfmt .",
     "lint": "oxlint",
     "test": "bun test",
-    "typecheck": "bun run predev && oxlint --type-aware --type-check",
+    "typecheck": "oxlint --type-aware --type-check",
     "prestart": "bun run build",
     "start": "./dist/server"
   },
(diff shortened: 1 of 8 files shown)


apps/api/src/lib/redact-provider.ts
```diff
@@ -1,13 +1,16 @@
-import type { TObject, TSchema, TUnion } from "@sinclair/typebox";
+type RedactableSchema = {
+  anyOf?: RedactableSchema[];
+  properties?: Record<string, Record<string, unknown>>;
+};
```
heiwen: Please revert this change. The point is to use the correct upstream types, so that if the types change, the typecheck actually complains. In the current version, you are just disabling the typecheck.

apps/api/package.json
```diff
@@ -16,7 +16,7 @@
     "format": "oxfmt .",
     "lint": "oxlint",
     "test": "bun test",
-    "typecheck": "bun run predev && oxlint --type-aware --type-check",
+    "typecheck": "oxlint --type-aware --type-check",
```
heiwen: This will break running typecheck from within the specific app (same for the others).

apps/api/package.json
```diff
@@ -31,6 +31,7 @@
   },
   "devDependencies": {
     "@hebo/typescript-config": "workspace:*",
+    "@sinclair/typebox": "^0.34.49",
```
heiwen: Please align version with exactly the same that Elysia requires

heiwen (approved):

PR #327, opened 2026-04-02
add cluster mode for Elysia services

turisanapo:
## Summary

- Adds a shared `serve()` utility in `packages/shared-api/lib/serve.ts` that handles `node:cluster` forking for multi-core usage.
- On Linux, forks one worker per CPU core (via Bun's `SO_REUSEPORT`). On macOS/Windows, runs single-process — zero config needed for local dev.
- All four Elysia services (`api`, `auth`, `gateway`, `mcp`) now use `serve()` instead of manual `.listen()` + `console.log()`.

> [!NOTE]
> ECS Fargate vCPUs are hyperthreaded, so `1 vCPU` gives `os.availableParallelism() = 2`.

## Test plan

- [x] `bun run typecheck` passes for all affected packages
- [x] `bun run dev` works on macOS (single-process, no forking)
- [x] Deploy to Linux staging and verify workers are forked (check logs for "primary forking N workers")

## Summary by CodeRabbit

* **Chores**
  * Unified server startup across API, Auth, Gateway, and MCP for consistent launch behavior and messaging.
  * Standardized startup output so services report resolved URLs and include process IDs when running multiple workers.
  * Introduced optional multi-worker support (configurable via a WORKERS env setting) to improve scalability in production on Linux.
  * Added example WORKERS entries to service env samples.

### DIFF
--- a/packages/shared-api/lib/serve.ts
+++ b/packages/shared-api/lib/serve.ts
@@ -9,10 +9,10 @@ export function serve(
   factory: () => AnyElysia,
   port: number,
   name: string,
-  options?: Record<string, unknown>,
+  options?: { idleTimeout?: number; workers?: number },
 ) {
   const workers =
-    Number(process.env.WORKERS) ||
+    options?.workers ??
     (IS_PRODUCTION && process.platform === "linux" ? os.availableParallelism() : 1);
 
   if (workers > 1 && cluster.isPrimary) {
@@ -20,7 +20,7 @@ export function serve(
     return;
   }
 
-  const app = factory().listen(options ? { port, ...options } : port);
+  const app = factory().listen({ port, idleTimeout: options?.idleTimeout });
   console.log(
     `🐵 ${name} running at ${app.server!.url}${workers > 1 ? ` (worker ${process.pid})` : ""}`,
   );
--- a/packages/shared-api/lib/serve.ts
+++ b/packages/shared-api/lib/serve.ts
@@ -7,8 +7,9 @@ import { IS_PRODUCTION } from "../env";
 
 export function serve(
   factory: () => AnyElysia,
-  port: Parameters<AnyElysia["listen"]>[0],
+  port: number,
   name: string,
+  options?: Record<string, unknown>,
 ) {
   const workers = IS_PRODUCTION && process.platform === "linux" ? os.availableParallelism() : 1;
 
@@ -17,7 +18,7 @@ export function serve(
     return;
   }
 
-  const app = factory().listen(port);
+  const app = factory().listen(options ? { port, ...options } : port);
   console.log(
     `🐵 ${name} running at ${app.server!.url}${workers > 1 ? ` (worker ${process.pid})` : ""}`,
   );
--- a/packages/shared-api/lib/serve.ts
+++ b/packages/shared-api/lib/serve.ts
@@ -1,19 +1,24 @@
 import cluster from "node:cluster";
 import os from "node:os";
 
-type ListenArg = string | number | Record<string, unknown>;
+import type { AnyElysia } from "elysia";
 
-// oxlint-disable-next-line typescript-eslint/no-explicit-any
-export function serve(factory: () => any, port: ListenArg, name: string) {
-  const workers = process.platform === "linux" ? os.availableParallelism() : 1;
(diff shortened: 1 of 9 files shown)


turisanapo:
@claude, how would you address coderabbit comments?

packages/shared-api/lib/serve.ts
```diff
@@ -0,0 +1,19 @@
+import cluster from "node:cluster";
+import os from "node:os";
+
+type ListenArg = string | number | Record<string, unknown>;
+
+// oxlint-disable-next-line typescript-eslint/no-explicit-any
```
heiwen: Please type properly. The factory should return an Elysia instance I suppose? That should then also address the remaining // oxlint-disable aspects.

packages/shared-api/lib/serve.ts
```diff
@@ -0,0 +1,19 @@
...
+export function serve(factory: () => any, port: ListenArg, name: string) {
+  const workers = process.platform === "linux" ? os.availableParallelism() : 1;
```
heiwen: Suggest:
1/ in DEVELOPMENT to also limit this to a single instance
2/ expose an environment variable on a service-by-service level that allows fine-tuning. (e.g. for when multiple services run on the same machine, to avoid overload)

packages/shared-api/lib/serve.ts
```diff
@@ -0,0 +1,19 @@
+import cluster from "node:cluster";
+import os from "node:os";
+
+type ListenArg = string | number | Record<string, unknown>;
```
heiwen: I suppose in our case we know the type exactly and don't have to support all these overloads?

packages/shared-api/lib/serve.ts
```diff
@@ -0,0 +1,19 @@
...
+
+// oxlint-disable-next-line typescript-eslint/no-explicit-any
+export function serve(factory: () => any, port: ListenArg, name: string) {
+  const workers = process.platform === "linux" ? os.availableParallelism() : 1;
+
+  if (workers > 1 && cluster.isPrimary) {
```
heiwen, line 18: The upstream documentation handles this differently, are you sure this is correct?

https://elysiajs.com/patterns/deploy

packages/shared-api/lib/serve.ts
```diff
@@ -1,19 +1,24 @@
...
+import type { AnyElysia } from "elysia";
 
-// oxlint-disable-next-line typescript-eslint/no-explicit-any
-export function serve(factory: () => any, port: ListenArg, name: string) {
-  const workers = process.platform === "linux" ? os.availableParallelism() : 1;
+import { IS_PRODUCTION } from "../env";
+
+export function serve(
+  factory: () => AnyElysia,
+  port: Parameters<AnyElysia["listen"]>[0],
```
heiwen: `Parameters<AnyElysia["listen"]>[0]` is unnecessarily complex. Are you passing a `string` or a `number` from the actual services?

packages/shared-api/lib/serve.ts
```diff
@@ -11,7 +11,9 @@ export function serve(
   name: string,
   options?: Record<string, unknown>,
 ) {
-  const workers = IS_PRODUCTION && process.platform === "linux" ? os.availableParallelism() : 1;
+  const workers =
+    Number(process.env.WORKERS) ||
```
heiwen: shared-api directly relying on process.env.WORKERS is a dependency violation. this should be passed as option to serve() instead.

packages/shared-api/lib/serve.ts
turisanapo: The upstream docs unconditionally fork because they assume cluster mode is always desired. Our version adds a `workers > 1` guard so the function gracefully falls back to single-process when:

- Running on macOS/Windows (no `SO_REUSEPORT`)
- Running on a single-core machine (`os.availableParallelism() === 1`)
- `WORKERS=1` is explicitly set

When `workers === 1`, the primary skips forking and runs the server directly — no overhead. When `workers > 1`, it matches the docs behavior: primary forks, workers run the server. The early `return` after forking mirrors the docs' if/else structure.

heiwen (approved):

PR #330, opened 2026-04-02
fix: add missing Azure OpenAI secrets to infra

turisanapo:
## Summary

- Declares `AzureOpenAiApiKey` and `AzureOpenAiResourceName` as SST secrets in `infra/stacks/env.ts`
- Adds them to the `llmSecrets` array so they're linked to the gateway service

## Problem

The gateway's `loadProviderSecrets()` calls `getSecret("AzureOpenAiApiKey")` and `getSecret("AzureOpenAiResourceName")`, but these were never declared as SST secrets. When `Resource[name]` throws, the fallback calls Bun's `secrets.get()` which crashes on distroless containers (no `libsecret`). This was introduced in `6a78b881` (March 21) but production is still running a March 19 build, so it hasn't surfaced yet. The next production deploy will crash without this fix.

## Summary by CodeRabbit

* **Chores**
  * Updated credential sourcing to use Foundry-named secrets for provider authentication.
  * Reordered provider entries to change provider priority order.
  * Added and exposed Foundry-related LLM secret entries in environment exports and secret loading.


infra/stacks/env.ts
```diff
@@ -21,6 +21,8 @@ export const smtpFrom = getSstSecret("SmtpFrom");
 
 // LLMs
 export const anthropicApiKey = getSstSecret("AnthropicApiKey");
+export const azureOpenAiApiKey = getSstSecret("AzureOpenAiApiKey");
+export const azureOpenAiResourceName = getSstSecret("AzureOpenAiResourceName");
```
heiwen: `AzureOpenAiApiKey` feels like a weird name. The service is Azure Foundry, isn't it?

turisanapo:
@claude, help me address heiwen's comment by renaming Azure OpenAI references to just `Foundry` to keep consistency with Amazon Bedrock references named just Bedrock, and Google Vertex named just Vertex.

heiwen (approved):

turisanapo:
@claude, can you keep the alphabetical order

PR #352, opened 2026-04-05
fix(gateway): Bedrock BYOK must not inherit AWS_SESSION_TOKEN (IRSA)

### LINKED ISSUE #351: Gateway: Bedrock BYOK fails in production with invalid security token when `AWS_SESSION_TOKEN` is set (IRSA)
### Summary

When Bedrock is configured via **Bring Your Own Key** (static `accessKeyId` / `secretAccessKey` stored per org), requests work locally but fail in production (e.g. EKS with IRSA) with:

`The security token included in the request is invalid`

### Root cause

`@ai-sdk/amazon-bedrock` **`createAmazonBedrock`** resolves SigV4 credentials by combining explicit access key options with **`AWS_SESSION_TOKEN` from the environment** when `sessionToken` is not passed in options (via `loadOptionalSetting` → `process.env.AWS_SESSION_TOKEN`).

On EKS/IRSA the pod always has **`AWS_SESSION_TOKEN`** (and related vars) for the **workload identity**, which does **not** match the **BYOK IAM user** key pair. The merged credentials are invalid for Bedrock SigV4.

References: [`createAmazonBedrock` credential assembly](https://github.com/vercel/ai/blob/main/packages/amazon-bedrock/src/bedrock-provider.ts) (same pattern as bundled `@ai-sdk/amazon-bedrock`).

### Steps to reproduce

1. Run the gateway **without** the `credentialProvider` workaround (i.e. pass only `accessKeyId` + `secretAccessKey` into `createAmazonBedrock`).
2. Configure Bedrock BYOK in the console with valid long-term IAM user keys.
3. Ensure the process has a **non-empty** `AWS_SESSION_TOKEN` in the environment (mimic IRSA):

   ```bash
   export AWS_SESSION_TOKEN=fake-token-for-repro
   ```

4. Call the gateway with a model that uses the BYOK Bedrock provider.

**Expected:** request succeeds using only the configured access key + secret.

**Actual:** Bedrock/AWS responds with *invalid security token*.

**Note:** After the fix in `apps/gateway/src/lib/provider.ts` (Bedrock `access-key` mode uses `credentialProvider: async () => ({ accessKeyId, secretAccessKey })`), env `AWS_SESSION_TOKEN` is **not** merged and the fake token no longer affects BYOK—so reproduction requires reverting that pattern or testing against a pre-fix build.

### Fix (in-repo)

Use `credentialProvider` returning **only** `{ accessKeyId, secretAccessKey }` so the AI SDK does not read `AWS_SESSION_TOKEN` for this path (`apps/gateway/src/lib/provider.ts`).

### Suggested upstream improvement

`createAmazonBedrock` could treat explicit static keys as authoritative and **not** fall back to `AWS_SESSION_TOKEN` unless `sessionToken` is explicitly desired, or document that BYOK deployments must avoid env session tokens / use `credentialProvider`.

### Environment

- Local dev: usually no `AWS_SESSION_TOKEN` → no bug.
- Production EKS + IRSA: `AWS_SESSION_TOKEN` set → bug without workaround.

turisanapo:
## Summary

Fixes #351.

`@ai-sdk/amazon-bedrock` builds SigV4 credentials from `accessKeyId` / `secretAccessKey` but still loads `AWS_SESSION_TOKEN` from the environment when no `sessionToken` option is set. On EKS/IRSA that merges the pod session token with BYOK IAM user keys and AWS rejects the request.

## Change

Use `credentialProvider: () => Promise.resolve({ accessKeyId, secretAccessKey })` for Bedrock `access-key` mode so only those fields are supplied (no env session token).

## Summary by CodeRabbit

* **Refactor**
  * Improved how Bedrock access-key authentication supplies credentials: the access-key flow now provides credentials via an asynchronous provider callback (while still validating required fields). This is an internal credential-handling change with no user-facing behavior changes.

### DIFF
--- a/apps/gateway/src/lib/provider.ts
+++ b/apps/gateway/src/lib/provider.ts
@@ -97,6 +97,8 @@ export function createProvider(slug: ProviderSlug, config: unknown): ProviderV3 
         case "access-key": {
           const { accessKeyId, secretAccessKey } = bedrockConfig;
           if (!accessKeyId || !secretAccessKey) return;
+          // credentialProvider: passing keys alone still merges AWS_SESSION_TOKEN from
+          // the process env (IRSA, etc.); see https://github.com/vercel/ai/issues/14136
           return withCanonicalIdsForBedrock(
             createAmazonBedrock({
               region,
--- a/apps/gateway/src/lib/provider.ts
+++ b/apps/gateway/src/lib/provider.ts
@@ -97,7 +97,6 @@ export function createProvider(slug: ProviderSlug, config: unknown): ProviderV3 
         case "access-key": {
           const { accessKeyId, secretAccessKey } = bedrockConfig;
           if (!accessKeyId || !secretAccessKey) return;
-          // See #351: credentialProvider avoids AWS_SESSION_TOKEN from IRSA + BYOK key mix-up.
           return withCanonicalIdsForBedrock(
             createAmazonBedrock({
               region,
--- a/apps/gateway/src/lib/provider.ts
+++ b/apps/gateway/src/lib/provider.ts
@@ -97,8 +97,12 @@ export function createProvider(slug: ProviderSlug, config: unknown): ProviderV3 
         case "access-key": {
           const { accessKeyId, secretAccessKey } = bedrockConfig;
           if (!accessKeyId || !secretAccessKey) return;
+          // See #351: credentialProvider avoids AWS_SESSION_TOKEN from IRSA + BYOK key mix-up.
           return withCanonicalIdsForBedrock(
-            createAmazonBedrock({ region, accessKeyId, secretAccessKey }),
+            createAmazonBedrock({
+              region,
+              credentialProvider: () => Promise.resolve({ accessKeyId, secretAccessKey }),
+            }),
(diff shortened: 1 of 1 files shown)


heiwen:
According to AI SDK documentation the most minimal fix would be to explicitly set `sessionToken: undefined`: https://ai-sdk.dev/providers/ai-sdk-providers/amazon-bedrock#provider-instance

turisanapo:
> According to AI SDK documentation the most minimal fix would be to explicitly set `sessionToken: undefined`: https://ai-sdk.dev/providers/ai-sdk-providers/amazon-bedrock#provider-instance


Good catch. I tried the docs approach (`sessionToken: undefined`) with a `AWS_SESSION_TOKEN` in the env; it **still** breaks. With the current SDK, `undefined` doesn’t stop `AWS_SESSION_TOKEN` from being picked up, so we’re using **`credentialProvider`** until a fixed `@ai-sdk/amazon-bedrock` lands ([vercel/ai#14137](https://github.com/vercel/ai/pull/14137)).

heiwen (approved):

PR #355, opened 2026-04-05
fix: prevent wrong model removal due to stale form state

### LINKED ISSUE #332: Failed to removing old model and causing new model being removed
https://github.com/user-attachments/assets/1e001f61-57ea-4cc6-b99a-9fb1bafa27bf

heiwen:
Fixes #332

Adds a `key` prop to the `useForm` hook in the model config form, derived from the server-provided models. This forces Conform to re-initialize when models update after revalidation, preventing stale `initialValue` from causing `form.reset()` to drop newly added models.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed form behavior in the models configuration so the form reliably updates and rebinds when switching between different model sets, ensuring controls reset to the correct model-specific values while preserving submit/reset behavior.

### DIFF
--- a/apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.models/form.tsx
+++ b/apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.models/form.tsx
@@ -67,7 +67,6 @@ export default function ModelsConfigForm({
   const fetcher = useFetcher<typeof clientAction>();
 
   const [form, fields] = useForm<ModelsConfigFormValues>({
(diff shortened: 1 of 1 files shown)


heiwen:
@claude key is not valid, please check confom documentation for the correct usage here

heiwen:
@claude how about using id property instead on the useForm?

heiwen:
@claude so id and key are effectively the same mechanisms just in different versions?

heiwen:
Closed in favor of https://github.com/8monkey-ai/hebo-platform/pull/357

PR #359, opened 2026-04-06
fix: guard lastResult with navigation.state and close dialog on remove confirm

### LINKED ISSUE #332: Failed to removing old model and causing new model being removed

heiwen:
Fixes #332

Two changes to fix the model removal bug:

1. Guard `lastResult` with `navigation.state === "idle"` per Conform's Remix integration docs. This ensures Conform only processes the action result (including `resetForm`) after parent route revalidation completes, preventing stale `initialValue` from causing `form.reset()` to drop newly added models.

2. Wrap the confirm Remove button in `DialogClose` so the dialog dismisses when the user confirms the removal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale form results from appearing during route/navigation transitions by ensuring form state only updates when navigation is idle.
  * Improved model list rendering to avoid unstable item identity during updates.
  * Improved the "Remove model" confirmation dialog so it reliably closes and shows operation progress while the removal runs.

### DIFF
--- a/apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.models/form.tsx
+++ b/apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.models/form.tsx
@@ -91,7 +91,7 @@ export default function ModelsConfigForm({
 
       {fields.models.getFieldList().map((model, index) => (
         <ModelCard
-          key={model.name}
+          key={model.key}
           model={model}
           agentSlug={agentSlug}
           branchSlug={branchSlug}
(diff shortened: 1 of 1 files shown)


heiwen:
@claude the close dialog issue is fixed. but no matter whether I stay on the page or even navigate back and forth to the page or hit reload, when I delete the agent, it's always the second one that is deleted, not the selected one. Why?

turisanapo (approved):

PR #362, opened 2026-04-06
unify API and Console schemas: replace TypeBox with Zod

### LINKED ISSUE #308: Unify API and Console schemas: replace dual TypeBox + Zod definitions with a single source of truth
## Context

Raised during review of #306 ([discussion](https://github.com/8monkey-ai/hebo-platform/pull/306#discussion_r2968821337)).

The API and Console currently define request/response schemas independently using different libraries:

- **API** (`apps/api/src`): TypeBox / Elysia `t.Object` / `t.Union` — used for request validation and response typing
- **Console** (`apps/console/app`): Zod — used for form validation (Conform) and mock data

This duplication spans across the entire codebase, not just providers:

| Domain | API (TypeBox) | Console (Zod) |
|--------|--------------|---------------|
| Providers | `providers/types.ts` — 22 TypeBox usages (config shapes, auth modes, slugs) | `configure.tsx` — provider config schemas, `clear.tsx` |
| Traces | `traces/types.ts` — 96 TypeBox usages (span details, messages, queries) | `mocks/routes/traces.ts` |
| Models | `branches.ts` — 12 TypeBox usages (model alias, routing) | `models/schema.ts` — model config form schema |
| Agents | `agents.ts` — 7 TypeBox usages | `agent.create/form.tsx` |
| Branches | `branches.ts` | `branches/create.tsx`, `branches/delete.tsx` |
| API Keys | — | `api-keys/create.tsx`, `api-keys/revoke.tsx` |
| Settings | — | `danger-zone.tsx`, `invite-schema.ts` |

Field names, enum values, and structural shapes must stay in sync manually. When one side changes without the other, it silently breaks.

This is a historical artifact from when Elysia didn't support [Standard Schema](https://standardschema.dev/). Now that Elysia supports it, we can unify on a single schema library.

## Proposed approach

1. Define shared schemas once using Zod (which implements Standard Schema) in a shared package (e.g. `packages/shared` or a dedicated `packages/schemas`).
2. Consume the shared schemas in both:
   - `apps/api` — Elysia validates using Standard Schema directly
   - `apps/console` — Zod schemas are used as-is for Conform form validation and mocks
3. Remove the TypeBox definitions from `apps/api/src`.

## Scope

This affects most API modules (`providers`, `traces`, `branches`, `agents`) and their corresponding console routes. Consider whether to:
- Migrate all modules at once, or
- Start with a smaller module (e.g. providers or agents) as a pilot and migrate the rest incrementally

## Related

- #306 — introduced the `authMode` discriminated union for providers, which amplified the duplication
- #300 — original issue for provider expansion

heiwen:
Replace dual TypeBox + Zod definitions with Zod as single source of truth, keeping all schemas in `apps/api`.

Closes #308

## Summary by CodeRabbit

* **Refactor**
  * Unified validation with Zod across API and frontend for consistent form validation and stricter inputs.
  * Consolidated provider and branch schemas to remove duplication and align frontend/backend shape handling.
  * Centralized Zod initialization to standardize validation error wording.

* **Chores**
  * Database schema defaults for audit fields tightened to improve data integrity.

* **Bug Fixes**
  * Simplified form validators: some custom hardcoded error messages were removed, normalizing validation feedback.


heiwen:
@claude typecheck is failing with 20+ errors

heiwen:
@claude what do you think about codex and coderabbit comments?

heiwen:
@claude address 1,2,3,4 &5

apps/api/src/modules/agents.ts
```diff
@@ -1,21 +1,19 @@
...
+import { agentsModelSchema } from "~api/generated/zod/schemas/variants/pure/agents.pure";
+import { agentsInputSchema } from "~api/generated/zod/schemas/variants/input/agents.input";
 import { prisma } from "~api/middlewares/prisma";
 
-const agents = t.Composite([agentsPlain, t.Partial(agentsRelations)], {
-  additionalProperties: false,
+const agentsPlain = agentsModelSchema.omit({ id: true });
+const agentsWithBranches = agentsPlain.extend({
+  branches: z.array(z.any()).optional(),
 });
```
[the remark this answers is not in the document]
heiwen: Please double check whether this has been resolved

heiwen:
@claude typecheck is failing now

heiwen:
@claude great, can you also rebase to main

heiwen:
@claude have a look at the one open coderabbit comment

apps/api/src/modules/agents.ts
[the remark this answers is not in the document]
heiwen: please check again

heiwen:
@claude what do you think about coderabbit's latest comments?

apps/api/src/db/utils.ts
```diff
@@ -1,21 +1,34 @@
...
 
-export function redactSensitiveValues<T>(schema: TSchema, value: T): T {
-  const obj = value as Record<string, unknown>;
+type ZodUnionMembers = readonly [z.ZodType, ...z.ZodType[]];
 
-  for (const variant of schema.anyOf ?? [schema]) {
-    for (const candidate of (variant as TUnion).anyOf ?? [variant]) {
-      const props = (candidate as TObject).properties ?? {};
+export function redactSensitiveValues<T>(schema: z.ZodUnion<ZodUnionMembers>, value: T): T {
+  const obj = value as Record<string, unknown>;
```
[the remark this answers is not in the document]
heiwen: The only caller is the Prisma result extension in prisma.ts:112, where value is always a ProviderConfig JSON column — always a non-null object from Prisma's JSON deserialization. Adding a guard for null/primitives is defensive against a scenario that can't happen given the current call site. If this utility were ever exported to a broader context it'd matter, but right now it's YAGNI.

heiwen:
@claude ok, address 2 and 3

heiwen:
@claude for 2, do we need the cast at all, we didn't have it before?

heiwen:
@claude resolve bun.lock conflicts

heiwen:
@claude
1/ for schemas where you don't use shared imports, can you keep the custom messages?
2/ in traces/route preserve previous semantics around default values

heiwen:
@claude why don't you keep setting the page / pageSize params on the zod schema instead? wouldn't that match the previous pattern better?

heiwen:
@claude make that change

heiwen:
@claude a handful of additional comments:
1/ Do we need `excludeFields` in the `models/{agents, branches,provider_config}` section of `zod-generator-config.json` or is it enough in the `variants`?
2/ In `ProviderSlug`, can we keep the previous pattern to use a z.enum, with Object.keys(…).map(…)?
3/ Instead of “textarea”, let’s call the meta() it “text", that's more isomorphic
4/ Suggest to ProviderConfigureSchema into the frontend, since that’s the only place where it is used
5/ Why are page / pageSize using a coerce() instead of optional? Same for from / to / timestamp?
6/ Is there no built-in type for agentsPlain / branchPlain / agentsWithBranches / agentsInclude that we can use from prisma-zod-generator instead of deriving our own? Same for branchesModelSchema.omit({ id: true, agent: true });
7/ ModelConfig in models/schema.ts, why are you extending it instead of using the original one?

Think about the right approach first and answer here. before implementing I'll confirm.

heiwen:
@claude
1/ the alternative is to add `id` on the variant level and remove them from the models, that would reduce complexity, wouldn’t it?
2/ got it, let’s just inline then
3/ let’s rename `textarea` meta to `text`
4/ move ProviderConfigureSchema back to the frontend, but keep import the base types from the backend, how does that sound?
5/ ok, keep coerce. But what about the missing optional?
6/ go ahead with your suggestion
7/ should we use the custom transform in the API as well then?

heiwen:
@claude in regards to item 7, I think it doesn't hurt to have that additional sanitization in the API and it simplifies the schema to be only in one place

heiwen:
@claude 
1/ I didn't ask to put BedrockIamRoleSchema, BedrockAccessKeySchema, VertexIdentityFederationSchema, .. back into the frontend. Just ProviderConfigureSchema which can then reference the types from the API.
2/ I still see the `branches` and `provider_config` in the `models` in zod-generator.config.json. Since they are already in `variants`, can't we remove them there?
3/ Let's make page / pageSize optional please, like they were before (keeping the default); that also allows us to remove the pageSize: 50 from the `traces/route`.
4/ Do we need this `const agentsPlain = agentsModelSchema`
5/ Is there a way to tell prisma-zod-generator to also generate objects with relations?
6/ Can we replace this with a generated type from prisma-zod-generator somehow: `branchPlain = branchesModelSchema.omit({ agent: true });` and `const branches = branchesModelSchema.omit({ agent: true });`

heiwen:
@claude
1/ if no relations are included in the schemas, why would we need `const branchPlain = branchesModelSchema.omit({ agent: true });`
2/ your model config in the frontend still extends ModelConfig, just reuse it from the API
3/ there are some typecheck errors

heiwen:
@claude I'm not happy with the relation handling yet. previously we didn't need casts. and excluding "agents" and "branches" specifically doesn't feel right either. what are the options here?

separately, I really like to keep the ModelConfig in the API, without transforming it. stop reverting that. if you hit issues, present me with options.

heiwen:
@claude:

1/ This is conceptually what we need: pureModelsIncludeRelations: true, result: enabled (or just omitted since default), pureModelsExcludeCircularRelations: true; wouldn't that work for us? Is there any way to get typed relations?

2/ I don't think you understand. I asked for the exact opposite. I don't want ModelConfig to extend in the frontend. I want the API to have the transforms().

heiwen:
@claude resolve conflicts

heiwen:
@claude what do you think about CodeRabbit comments ?

heiwen:
@claude fix 1,3,4 and 9

apps/console/app/mocks/routes/providers.ts
```diff
@@ -41,12 +25,13 @@ export const providerHandlers = [
...
+    const o = ProviderSchema.options.find((option) => option.shape.slug.value === params.slug);
 
     db.providers.delete((q) => q.where({ slug: params.slug }));
 
     const provider = await db.providers.create({
       slug: params.slug,
-      name: SUPPORTED_PROVIDERS[params.slug].name,
+      name: o!.shape.name.value,
       config: body,
     });
```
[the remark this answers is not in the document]
heiwen: This is a mock, unnessecary.

apps/api/src/modules/traces/types.ts
```diff
@@ -1,134 +1,124 @@
...
+  })
+  .loose();
+
+const GenAIMessagePartSchema = z.union([
+  TextPartSchema,
+  ReasoningPartSchema,
+  ToolCallPartSchema,
+  ToolCallResponsePartSchema,
+  GenericPartSchema,
 ]);
```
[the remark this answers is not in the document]
heiwen: Existing issue and data is validated upstream alrdy.

apps/console/app/routes/_shell.agent.create/route.tsx
```diff
@@ -26,7 +27,7 @@ export async function clientAction({ request }: Route.ClientActionArgs) {
 
   if (result.error?.status === 409)
     return submission.reply({
-      fieldErrors: { agentName: [parseError(result.error.value).message] },
+      fieldErrors: { name: [parseError(result.error.value).message] },
     });
 
   return redirect(`/agent/${result.data!.slug}/branch/${result.data!.branches![0].slug}`);
```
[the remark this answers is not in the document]
heiwen: Unrelated to this PR

apps/console/app/routes/_shell.agent.$agentSlug.branches/create.tsx
```diff
@@ -46,11 +40,11 @@ type CreateBranchProps = {
 
 export default function CreateBranch({ branches }: CreateBranchProps) {
   const fetcher = useFetcher<typeof clientAction>();
-  const [form, fields] = useForm<BranchCreateFormValues>({
+  const [form, fields] = useForm<BranchCreate>({
     lastResult: fetcher.state === "idle" ? fetcher.data?.submission : undefined,
     constraint: getZodConstraint(BranchCreateSchema),
     defaultValue: {
-      sourceBranchSlug: branches[0].slug,
+      source_branch_slug: branches[0].slug,
```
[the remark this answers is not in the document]
heiwen: That's impossible to happen by backend logic

heiwen:
@claude can you do a clean rebase to main

turisanapo:
@coderabbitai review

apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.providers/route.tsx
```diff
@@ -27,9 +27,9 @@ export async function clientAction({ request }: Route.ClientActionArgs) {
 
       let provider;
       try {
-        provider = await api.providers({ slug: submission.value.slug }).config.put({
-          ...submission.value.config,
-        });
+        provider = await api
+          .providers({ slug: submission.value.slug })
+          .config.put(submission.value.config as Record<string, unknown>);
```
turisanapo, line 32: Why not use `ProviderConfig` here?

apps/api/src/db/json-types.d.ts
turisanapo, line 1: Can this be a regular `.ts` file instead of `.d.ts`? `declare global` works in both.

apps/api/prisma/schema.prisma
```diff
@@ -7,97 +7,68 @@ generator client {
...
-  /// @prismabox.input.hide
   branches branches[]
 
-  // Common audit fields across entities
-  /// @prismabox.input.hide
-  created_by      String
-  /// @prismabox.input.hide
+  // Common audit fields — injected by Prisma extension.
+  // @default("") keeps create inputs optional, results non-nullable.
+  created_by      String    @default("")
```
turisanapo, line 27: Are there other options than introducing `@default("")`? This applies to `created_by`, `updated_by`, and `organization_id` across all three tables.

These are critical audit/tenancy fields that should never be empty strings. I understand the application always injects them, but adding a default to the schema means the database can no longer enforce their presence.

What about a type-level workaround (i.e., specific casts)? It would keep the DB constraint intact and avoid the migration too.

apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.providers/route.tsx
heiwen: It's a complicated Type inference issue with zod discriminated unions that I wasn't able to figure out even after 4h with Claude Opus.

apps/api/src/db/json-types.d.ts
heiwen: It shouldn't, because it really only contains types. That's what .d.ts is meant for.

apps/api/prisma/schema.prisma
heiwen: The only other option is to keep the manual casts in the agent / branch service files. 

But that's harder to work with for agents (and humans).

As long as the extension code is correct, the fields will always be set.

turisanapo: I would keep the manual casts.

1/ The required casts are only 4.

2/ A mandatory audit/tenancy field with `@default("")` can be read as optional at the schema level. With casts, the schema stays, as it should, the source of truth.

3/ This PR is a TypeBox-to-Zod replacement. It shouldn't concern the DB or migrations.

heiwen: Disagree

1/ It makes working with agents way harder, they run into very long loops trying to figure this out 

2/ it's properly guarded by the existing client extension

heiwen: Plus: the previous cast has the issue that any other required field might actually be missed, without an way to detect that. so it's actually making the same problem way bigger

Keep in mind: casts by themselves are insecure and should be avoided

turisanapo: Agree, I tried to find a solution that would accommodate both no casts and no changes on the schema. My concerns remain though:

1. Does it make sense to change the schema because we are changing how we validate the contract between client and server?

2. What happens if we decide to change again, or zod or prisma release something that would simplify the type? Does it make sense to change the DB again?

heiwen: Nothing to do with Zod or Typebox. The issue is purely a Prisma issue. It surfaces in this PR because it impacts the general ability of agents to work with the codebase and with that our productivity as a team.

turisanapo (approved):

PR #76, opened 2025-09-25
HEB-122 agent configuration

Waazify:
## Summary by CodeRabbit

- New Features
  - Configure models per agent branch with validated add/edit/remove, per-item save, copy-to-clipboard, loading states and success toasts.
  - Persistent branch updates with server-backed PATCH and structured success/error feedback.

- UI
  - Sidebar navigation grouped; “Models” link enabled when an agent is selected, disabled otherwise.
  - Collapsible model rows and animated row reveal for smoother editing.

- Improvements
  - Branch listings scoped to the selected agent.

- Chores
  - Added new UI dependencies for collapsible and radio controls.

PR #390, opened 2026-04-15
chore: use forked @elysiajs/opentelemetry with pending upstream patches

### LINKED ISSUE #389: chore: use forked @elysiajs/opentelemetry with pending upstream patches
## Context

We have two open PRs against `elysiajs/opentelemetry` that add features we need now:

- [#74 — safe-by-default span attributes: header allow-lists, URL redaction, body opt-in](https://github.com/elysiajs/opentelemetry/pull/74)
- [#76 — `http.server.request.duration` histogram metric](https://github.com/elysiajs/opentelemetry/pull/76)

Until they are merged and released, we maintain a **`fork/main`** branch on [turisanapo/opentelemetry](https://github.com/turisanapo/opentelemetry/tree/fork/main) that merges both PRs on top of upstream `main` and includes the pre-built `dist/` so it can be consumed as a git dependency without a build step.

## What to change

In the root `package.json`, update the Bun workspace catalog entry:

```diff
- "@elysiajs/opentelemetry": "^1.4.11",
+ "@elysiajs/opentelemetry": "github:turisanapo/opentelemetry#fork/main",
```

Then run:

```bash
bun install
```

All five workspaces (`apps/api`, `apps/auth`, `apps/gateway`, `apps/mcp`, `packages/shared-api`) reference `"@elysiajs/opentelemetry": "catalog:"`, so the single catalog change covers everything.

## How to validate

1. **Install succeeds** — `bun install` resolves the git dependency without errors.
2. **Build passes** — `bun run build` completes for all workspaces.
3. **Types work** — `bun run typecheck` passes. The new options (`headersToSpanAttributes`, `spanUrlRedaction`, `recordBody`) should be visible in autocomplete.
4. **Runtime smoke test** — start a local dev server (`bun run dev`) and confirm:
   - Traces still appear in your collector/backend as before.
   - If you pass `headersToSpanAttributes: { requestHeaders: ['x-request-id'] }`, only that header shows up on spans.
   - If you configure a `MeterProvider`, you see `http.server.request.duration` histogram data points.
5. **Existing tests** — `bun run test` passes with no regressions.

## Reverting (once upstream merges)

When both PRs land in a published version, revert the catalog entry to the registry version:

```diff
- "@elysiajs/opentelemetry": "github:turisanapo/opentelemetry#fork/main",
+ "@elysiajs/opentelemetry": "^<new-version>",
```

turisanapo:
Switch the workspace catalog entry from the registry version to the turisanapo/opentelemetry fork (fork/main branch) which includes PRs #74 (safe-by-default span attributes) and #76 (http.server.request.duration histogram metric).

Closes #389

## Summary by CodeRabbit

* **Chores**
  * Updated a dependency to use an alternative source.
* **Bug Fixes**
  * Telemetry/span handling updated to use explicit allowlists for which headers become span attributes; span processor instantiation simplified to avoid per-span mutations.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -259,7 +259,7 @@
     "@elysiajs/cors": "^1.4.1",
     "@elysiajs/eden": "^1.4.8",
     "@elysiajs/openapi": "^1.4.14",
-    "@elysiajs/opentelemetry": "^1.4.11",
+    "@elysiajs/opentelemetry": "github:turisanapo/opentelemetry#fork/main",
     "@opentelemetry/api": "^1.9.0",
     "@opentelemetry/instrumentation": "^0.214.0",
     "@opentelemetry/instrumentation-pg": "^0.66.0",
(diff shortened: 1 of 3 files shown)


turisanapo:
@claude, what do you think of codeRabbit's comment?

heiwen (commented):
Should this include using the new features instead of current workarounds (e.g. header include list / redaction logic)?

turisanapo:
@claude, please help me remove the workaround where we explicitly redact sensitive headers.

Since now every header will be excluded, can you please help create a list of headers that still make sense to track?

heiwen (approved):

turisanapo:
@heiwen, based on your experience with the opentelemetry semantic conventions for generative ai, are there other headers we should whitelist?

heiwen:
@claude what do you think about coderabbits comment about "Drop referer and x-forwarded-for from the span header allow-list"? if you disagree skip, If you agree, implement.

heiwen:
@claude resolve the merge conflicts

heiwen (approved):

heiwen:
Might be worth doing another deletion of the `traces` table after this, to remove the number of columns introduced from bot's spamming with random headers.

Issue #333, opened 2026-04-03
Gateway: request body transfer adds 600ms+ latency on large conversation payloads


turisanapo:
## Problem

The gateway's `await ctx.request.json()` call blocks the handler until the full HTTP body has been received and parsed. For production Vertex requests, where 48% of payloads are 100–500 KB (conversation history), this introduces **600 ms at p50** and **1,055 ms at p95** of pure body transfer latency before any gateway logic executes. This is the dominant contributor (**98.1%**) to the observed pre-fetch overhead.

### Methodology

We instrumented the chat handler with `addSpanEvent()` calls at each phase boundary. By subtracting nanosecond timestamps of `hebo.handler.started` from `hebo.request.deserialized`, we isolate the time spent in `await ctx.request.json()` from any downstream processing. Body sizes come from the `content-length` HTTP header captured on the `POST /v1/chat/completions` span.

**Sample:** 3,514 Google Vertex streaming requests across 4 ECS tasks over a 4-hour window (2026-04-02).

### 1. Handler phase breakdown

| Phase | p50 | p90 | p95 | % of total p50 |
|-------|-----|-----|-----|----------------|
| **A. Body transfer** (`handler.started` → `request.deserialized`) | **599 ms** | 999 ms | 1,055 ms | **98.1%** |
| B. Gateway processing (`request.deserialized` → `ai-sdk.started`) | 5 ms | 8 ms | 10 ms | 0.8% |
| C. Auth + AI SDK overhead (`ai-sdk.started` → `fetch.start`) | 5 ms | 152 ms | 165 ms | 0.8% |
| **TOTAL pre-fetch** | **611 ms** | 1,038 ms | 1,094 ms | 100% |

### 2. Body size directly predicts deserialization time

| Body Size | Count | Deserialize p50 | Deserialize p90 |
|-----------|-------|-----------------|-----------------|
| < 10 KB | 910 | 0 ms | 0 ms |
| 10–50 KB | 453 | 202 ms | 414 ms |
| 50–100 KB | 228 | 410 ms | 433 ms |
| **100–500 KB** | **1,581** | **636 ms** | **852 ms** |
| 500 KB+ | 292 | 1,053 ms | 1,257 ms |

The linear relationship confirms the delay is network transfer time, not CPU-bound JSON parsing.

### 3. Content-Length distribution (production traffic)

| Bucket | Count | % |
|--------|-------|---|
| < 1 KB | 164 | 4.2% |
[...]

Issue #289, opened 2026-03-19
Cloud: Introduce SST Router in front of API services


heiwen:
# Infra: Introduce SST Router in front of API services

## Background

All three services (`api`, `auth`, `gateway`) are ECS Fargate containers behind Application Load Balancers in a single AWS region. TLS is terminated at each ALB. Every browser request travels the full round-trip from the client's PoP to `us-east-1` (or wherever the cluster lives).

[SST's Router](https://sst.dev/docs/component/aws/router) is a CloudFront distribution with path-based routing rules that sit in front of the ALBs:

```ts
export const router = new sst.aws.Router("HeboRouter", {
  routes: {
    "/api/*":     heboApi.url,
    "/auth/*":    heboAuth.url,
    "/gateway/*": heboGateway.url,
  },
  domain: isProduction ? "api.hebo.ai" : `api.${normalizedStage}.hebo.ai`,
});
```

---

### 1. TLS termination at the edge
Today, TLS handshakes happen at the ALB — a single-region endpoint. CloudFront terminates TLS at the nearest of ~600 global PoPs. For users outside `us-east-1`, this alone cuts connection setup time significantly (a TLS 1.3 handshake from Europe or Asia to a US ALB adds 150–250 ms; terminating it locally drops that to <10 ms).

### 2. TCP connection reuse across requests
CloudFront maintains persistent, multiplexed connections to the ALB origins over HTTP/2. Browser clients establish one TLS session to the nearest edge PoP and reuse it across all requests to `api`, `auth`, and `gateway`. Without the router, the browser opens three separate TCP+TLS connections (one per subdomain).

### 3. Faster DNS resolution
CloudFront domains resolve via Anycast to the nearest PoP. Per-service ALB DNS records resolve to a regional endpoint. Clients in non-US regions see meaningfully better TTFB for the first request.

### 4. AWS Shield Standard + WAF attachment point
CloudFront is the natural place to attach AWS WAF rules and gets Shield Standard free. Currently, to add WAF protection, it would need to be attached to three separate ALBs. See [WAF section](#waf-configuration) below for specifics.

### 5. Centralized TLS policy
The per-service `ELBSecurityPolicy-TLS13-1-2-2021-06` transform can be removed from each stack. TLS version policy is enforced once at the router.

---

### Streaming / SSE (high priority)
The gateway uses SSE for streaming LLM responses and has a 5-minute ALB idle timeout. CloudFront supports streaming but requires explicit configuration:
[...]

PR #399, opened 2026-04-15
Add per-service CloudFront routers (SST Router)

turisanapo:
## Summary

- Place an `sst.aws.Router` (CloudFront distribution) in front of each service ALB for edge TLS termination, TCP connection pooling, and future WAF support
- Remove `loadBalancer.domain` and HTTPS listeners from all services — ALBs now serve HTTP-only on auto-generated ELB DNS names
- Internal service-to-service calls (`AUTH_URL`) route directly to ALB DNS, bypassing CloudFront
- Gateway router uses `readTimeout: "60 seconds"` for LLM first-token latency (SSE heartbeats keep the connection alive after that)
- Console `VITE_*` env vars now reference Router URLs
- **Unify env vars**: replace `API_URL`/`GATEWAY_URL` with `BASE_URL` across all services; derive `ROOT_DOMAIN` in `shared-api` for CORS and cookie domain

## Architecture

```
Browser → CloudFront (TLS at edge) → ALB (HTTP) → ECS
                                          ↑
Service-to-service calls go here directly ─┘
```

Each service gets its own CloudFront distribution (Option B from #289):
- `auth.hebo.ai` → `HeboAuthRouter` → Auth ALB
- `api.hebo.ai` → `HeboApiRouter` → API ALB
- `gateway.hebo.ai` → `HeboGatewayRouter` → Gateway ALB
- `mcp.hebo.ai` → `HeboMcpRouter` → MCP ALB

## Key design decisions

- **`router.route()` over inline `routes`**: The inline `routes` constructor is deprecated and has `originReadTimeout` hardcoded to 20s. `router.route()` supports configurable `readTimeout` up to 60s.
- **HTTP-only ALBs**: CloudFront terminates TLS at the edge and talks HTTP to the ALB (same VPC, same region). No need for HTTPS on the ALB.
- **Internal traffic bypasses CloudFront**: `AUTH_URL` for API and Gateway uses `service.nodes.loadBalancer.dnsName` (raw ALB DNS) instead of the public domain.
- **No WAF initially**: WAF will be added after collecting baseline per-IP request rate data from GreptimeDB.

### Env var redesign

Moving `AUTH_URL` to the internal ALB broke CORS and cookie domain derivation — `getRootDomain("http://<alb>.amazonaws.com")` returned `amazonaws.com` instead of `hebo.ai`. The same issue affected cross-subdomain cookies via `shared-api/lib/better-auth.ts`.

The fix introduces a clear separation of concerns:

| Env var | Purpose | Cloud value | Local/self-hosted |
|---------|---------|-------------|-------------------|
| `BASE_URL` | OpenAPI docs, Better Auth `baseURL` | Set by infra (`https://api.hebo.ai`) | Defaults to `http://localhost:PORT` per app |
| `ROOT_DOMAIN` | CORS origin check, cookie domain | Derived from `BASE_URL` → `hebo.ai` | `undefined` → permissive (correct) |
| `AUTH_URL` | Service-to-service auth calls | Internal ALB DNS | `http://localhost:8523` (default) |

- `ROOT_DOMAIN` is computed once in `shared-api/env.ts` from `process.env.BASE_URL` and imported by `cors.ts` and `better-auth.ts`
- Each app defines its own `BASE_URL` locally with a port-specific fallback — no shared-api export needed
- Self-hosted Docker Compose and local dev require zero new env vars

## Requirements checklist (from #289)

| # | Requirement | Status | Notes |
|---|-------------|--------|-------|
| 1 | TLS termination at the edge | Done | CloudFront terminates TLS at ~600 global PoPs |
| 2 | TCP connection reuse across requests | Done | CloudFront maintains persistent HTTP/2 pools to ALB origins |
| 3 | Faster DNS resolution (Anycast) | Done | CloudFront domains resolve via Anycast |
| 4 | AWS Shield Standard | Done | Free and automatic on every CloudFront distribution |
| 5 | Centralized TLS policy | Done | TLS listener transforms removed from all services |
| 6 | Caching disabled for API routes | Done | SST Router defaults to `defaultTtl: 0`, `minTtl: 0` — verified from source |
| 7 | Authorization/Cookie header forwarding | Done | `AllViewerExceptHostHeader` policy forwards both |
| 8 | Internal traffic bypasses CloudFront | Done | `AUTH_URL` uses raw ALB DNS via `nodes.loadBalancer.dnsName` |
| 9 | Path prefix handling (no rewriting needed) | Done | Option B — each service has its own Router, no prefix stripping |
| 10 | Streaming/SSE support | Done | Gateway Router has `readTimeout: "60 seconds"`, SSE heartbeats every 20s keep connections alive |
| 11 | Dev environment story | Done | Option B means no local proxy needed — services keep their own ports |
| 12 | WAF | Deferred | Follow-up PR after collecting baseline per-IP rate data from GreptimeDB |
| 13 | CORS with internal AUTH_URL | Done | `ROOT_DOMAIN` derived from `BASE_URL`, not `AUTH_URL` |

## Workaround for SST bug

SST's CloudFront Function `setUrlOrigin` has a bug for HTTP origins (sst/sst#6742): it deletes `customOriginConfig` but leaves `originAccessControlConfig`, causing CloudFront to default to HTTPS:443 and return 502. A `$transform(aws.cloudfront.Function, ...)` in `sst.config.ts` patches the generated code at deploy time to set explicit HTTP origin config.

## Test plan

- [x] Deploy to preview via `preview deploy` label
- [x] Console loads and authenticates (cookie forwarding)
- [x] API CRUD operations work (Authorization header forwarding)
- [x] Gateway SSE streaming end-to-end (chat completion with streaming)
- [x] Verify `X-Amz-Cf-Id` header present on browser requests (CloudFront is in the path)
- [x] Verify internal auth calls don't have `X-Amz-Cf-Id` (ALB direct)
- [x] CORS preflight returns correct `Access-Control-Allow-Origin`

Depends on: #398
Ref: #289, #333

## Summary by CodeRabbit

* **Chores**
  * Standardized service env vars: replaced API_URL/AUTH_URL/GATEWAY_URL with BASE_URL across services and examples
  * Simplified load balancers: removed HTTPS redirects and consolidated to single HTTP forward rules
  * Added a CloudFront origin patch and wired the new routing into deployment startup

* **New Features**
  * Centralized hostname/ALB URL helpers and exported routers for api/auth/gateway/mcp
  * Exposed ROOT_DOMAIN from shared configuration and moved init-process toggle into infra helpers


infra/stacks/gateway.ts
```diff
@@ -42,23 +41,19 @@ const heboGateway = new sst.aws.Service("HeboGateway", {
...
     PORT: gatewayPort,
     ...(heboGreptime ? { GREPTIME_HOST: heboGreptime.service } : {}),
   },
   loadBalancer: {
-    domain: gatewayDomain,
-    rules: [
-      { listen: "80/http", redirect: "443/https" },
-      { listen: "443/https", forward: `${gatewayPort}/http` },
-    ],
+    rules: [{ listen: "80/http", forward: `${gatewayPort}/http` }],
```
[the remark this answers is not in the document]
turisanapo: Valid point — the ALBs are internet-facing by default and bypassing CloudFront via the raw ALB DNS is possible.

The pragmatic fix is restricting each ALB's security group to CloudFront's managed prefix list (`com.amazonaws.global.cloudfront.origin-facing`) via `transform.loadBalancerSecurityGroup`. This is the standard AWS pattern for CloudFront → public ALB and doesn't require private subnets or NAT gateways.

That said, this is a pre-existing condition (ALBs were already public before this PR) and orthogonal to the Router introduction. I'll track it as a follow-up alongside the WAF work in #289, since both are about hardening the edge layer.

infra/stacks/router.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const apiRouter = new sst.aws.Router("HeboApiRouter", {
+  domain: hostname("api"),
+});
+apiRouter.route("/", albUrl(heboApi));
+
+export const gatewayRouter = new sst.aws.Router("HeboGatewayRouter", {
+  domain: hostname("gateway"),
+});
+gatewayRouter.route("/", albUrl(heboGateway), { readTimeout: "60 seconds" });
```
heiwen: Why a specific `readTimeout` here but not on the others?

apps/api/.env.example
```diff
@@ -1,5 +1,5 @@
-API_URL=http://localhost:8521
 AUTH_URL=http://localhost:8523
+BASE_URL=http://localhost:8521
```
heiwen: Should BASE_URL really have a port, that feels inconsistent with a separate PORT variable?

heiwen (changes requested):

infra/stacks/gateway.ts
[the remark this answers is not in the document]
heiwen: I always check how SST handles this in their own `console` application for a blueprint: https://github.com/anomalyco/console

infra/stacks/router.ts
turisanapo: The default `readTimeout` is 20s. The gateway uses SSE with heartbeats every ~20s, so the default leaves no margin. 60s gives headroom without downside. The other services are normal request/response and well within 20s.

turisanapo: Also worth noting: `router.route()` caps `readTimeout` at 60s (CloudFront hard limit). For streaming, this is fine — each SSE chunk resets the idle timer. For non-streaming completions (`stream: false`), any response exceeding 60s would 504.

apps/api/.env.example
turisanapo: Agreed — the port was redundant. Simplified the fallback to `http://localhost` (no port) and updated the `.env.example` files to match.

heiwen: Don't see the actual change

turisanapo: Updated.

heiwen: The OpenAPI config still needs the full version with the PORT though

turisanapo: I had the same reasoning. The tradeoff is between duplicating the port in `.env.example` and having OpenAPI docs correctly configured in local dev. I decided to simplify the example since having Swagger working locally seems marginal. In production, we still apply the correct configuration.

turisanapo: Updated. Ports for services in production are neither `80` nor `443`. The condition is based on `localhost` being part of `BASE_URL`.

heiwen: 'Localhost' doesn't seem to be the right concept here to make that decision.

turisanapo: Right, relying only on localhost is brittle. I updated the code by including a more robust helper.

heiwen: Port 80 should also not be nessecary in the URL?

turisanapo: Updated the helper to prevent appending ports `80` and `443`.

heiwen: The ports should only be ignored if they match the protocol. 

Port is already a number in the places where getServerUrl is called.

turisanapo: Updated to avoid the cast and to ignore the port only when matching the protocol.

heiwen: The function to maybe attach the port to the url can be simplified

```
function withPort(base: string, port?: number) {
  const url = new URL(base);
  if (port) url.port = String(port);
  return url.href;
}
```

(this automatically omits the port if it matches the protocol)

turisanapo: Updated.

heiwen: `getServerUrl` remains very ambiguous, not really describing what's happening here.

turisanapo: Updated.

heiwen (approved):

infra/stacks/router.ts
heiwen, line 1: We generally organize our stacks by `domain`, not by `technical underlying service`. Based on that, the routers should simply move into each of the existing services.

heiwen (approved):

PR #404, opened 2026-04-16
improve auth failure observability for 401 debugging

### LINKED ISSUE #397: Transient 401s on gateway
## Summary

Genuine API consumers are intermittently receiving `401 Unauthorized` responses from the gateway despite sending valid API keys. The issue is transient — the same client with the same key gets both 200s and 401s in quick succession during bursts of concurrent requests.

## Root Cause (confirmed)

The auth service's `POST /internal/verify-api-key` handler stops matching during concurrent request bursts. The auth service receives the request (visible as `POST /*` in traces) but the route-specific handler never executes — no DB lookups, no API key validation. It returns a default 200 with a null/empty body, which the gateway interprets as an invalid key and returns 401.

**Root cause hypothesis (unconfirmed):** The request body becomes unavailable when it reaches `better-call`'s router under concurrent load. `better-call`'s `getBody()` returns `undefined` when `request.body` is null, Zod body validation fails, and the handler is skipped. We have not directly observed `request.body === null` in production yet, and there is no known upstream issue that exactly matches this scenario.

## Impact (from GreptimeDB log + trace analysis)

**812 requests** received spurious 401s over a ~21-hour incident window (Apr 13):

| Hour (UTC)       | Auth Failures | ~Est. Requests | Failure Rate |
|------------------|---------------|----------------|--------------|
| Apr 13 05:00     | 61            | ~11,238        | 0.54%        |
| Apr 13 06:00     | 87            | ~10,868        | 0.80%        |
| Apr 13 07:00     | 1             | ~2,750         | 0.04%        |
| **Apr 13 08:00** | **354**       | **~3,588**     | **9.87%**    |
| Apr 13 09:00     | 147           | ~2,303         | 6.38%        |
| Apr 13 10:00     | 106           | ~1,820         | 5.82%        |
| Apr 13 11:00     | 45            | ~1,643         | 2.74%        |
| Apr 13 15:00     | 1             | ~1,287         | 0.08%        |
| Apr 14 03:00     | 10            | ~1,772         | 0.56%        |
| **Total**        | **812**       | **~37,994**    | **2.14%**    |

A separate recurrence on Apr 15 (28 failures in a 10-second burst at 08:44 UTC) provided correlated trace data — see [follow-up comment](#issuecomment-4256607333).

## Observability Gap

We **cannot identify which consumers were affected** in the Apr 13 incident because:

1. The log entry is completely bare: `{"msg":"Authentication failed or no credentials provided"}` — no client IP, user agent, API key prefix, or error detail
2. `log_attributes` is empty `{}`
3. Trace span data from that period rotated out of GreptimeDB (traces have shorter retention than logs)
4. The `error` property from the `verifyApiKey` call is silently discarded by the destructuring pattern

### 1. Structured error logging (observability)

Add client context and the actual error to the auth failure log so incidents are diagnosable:

```typescript
if (!organizationId || !userId) {
  logger.info({
    clientIp: request.headers.get("x-forwarded-for") || "unknown",
    userAgent: request.headers.get("user-agent")?.slice(0, 100),
    hasAuthHeader: !!authHeader,
    hasCookieHeader: !!cookieHeader,
    verifyResult: authHeader ? (result === null ? "null_response" : "invalid") : undefined,
  }, "Authentication failed or no credentials provided");
}
```

Also capture the `error` from the `verifyApiKey` response:

```typescript
const { data: result, error: verifyError } = await authClient.internal.verifyApiKey(...);
if (verifyError) {
  logger.warn({ error: verifyError }, "API key verification request failed");
}
```

### 2. Add `onRequest` hook to auth service to confirm hypothesis

Log `request.body` state before `better-call` processes it:

```typescript
app.onRequest(({ request }) => {
  if (request.method === "POST") {
    logger.info({ bodyIsNull: request.body === null, url: request.url }, "auth:request-body-check");
  }
});
```

### 3. Retry on null response (resilience)

When `verifyApiKey` returns `data: null` without an error, retry once before giving up:

```typescript
let { data: result, error: verifyError } = await authClient.internal.verifyApiKey(...);
if (!result && !verifyError) {
  ({ data: result, error: verifyError } = await authClient.internal.verifyApiKey(...));
}
```

### 4. Short-lived LRU cache for valid API keys (performance + resilience)

Cache successful API key verifications in the gateway for a short TTL (e.g. 30–60s) to reduce auth service calls and eliminate the concurrency window entirely for recently-verified keys.

### 5. Increase GreptimeDB trace retention

Current trace retention is extremely short (~20 minutes observed). Logs survive ~72+ hours. Align trace retention with logs so future incidents can be fully correlated.

turisanapo:
Closes #397 partially — adds observability to diagnose the root cause.

## Changes

- Capture `verifyError` from `authClient.internal.verifyApiKey` in gateway auth middleware
- Wire better-auth’s internal logger into the OTEL pipeline

## Summary by CodeRabbit

* **Chores**
  * Introduced a shared, consistent logger across services and apps for unified structured logs and telemetry.
  * Consolidated log level and severity parsing to derive both together for predictable runtime logging.

* **Bug Fixes**
  * Improved error and warning emissions during API key verification so failures are logged and surfaced more reliably.

### DIFF
--- a/apps/auth/src/better-auth.ts
+++ b/apps/auth/src/better-auth.ts
@@ -52,7 +52,7 @@ export const auth = betterAuth({
   logger: {
     level: LOG_LEVEL as "debug" | "info" | "warn" | "error",
     log: (level, message, ...args) => {
-      authLogger[level]({ source: "better-auth", args }, message);
+      authLogger[level]({ args }, message);
     },
   },
   accountLinking: {
--- a/apps/auth/src/better-auth.ts
+++ b/apps/auth/src/better-auth.ts
@@ -52,8 +52,7 @@ export const auth = betterAuth({
   logger: {
     level: LOG_LEVEL as "debug" | "info" | "warn" | "error",
     log: (level, message, ...args) => {
-      const fn = authLogger[level] ?? authLogger.info;
-      fn({ source: "better-auth", args: args.length > 0 ? args : undefined }, message);
+      authLogger[level]({ source: "better-auth", args }, message);
     },
   },
   accountLinking: {
--- a/apps/auth/src/better-auth.ts
+++ b/apps/auth/src/better-auth.ts
@@ -4,8 +4,10 @@ import { betterAuth } from "better-auth/minimal";
 import { emailOTP, organization } from "better-auth/plugins";
 
 import { createPrismaAdapter } from "@hebo/shared-api/db/postgres";
-import { AUTH_SECRET, AUTH_URL, LOG_LEVEL } from "@hebo/shared-api/env";
+import { AUTH_SECRET, AUTH_URL, LOG_LEVEL, LOG_SEVERITY } from "@hebo/shared-api/env";
 import { COOKIE_CONFIG } from "@hebo/shared-api/lib/better-auth";
+import { createOtelLogger } from "@hebo/shared-api/lib/otel";
+import { createPinoOtelAdapter } from "@hebo/shared-api/utils/otel-pino";
 import { getSecret } from "@hebo/shared-api/utils/secret";
 import { getRootDomain } from "@hebo/shared-api/utils/url";
 
@@ -24,6 +26,8 @@ const prisma = new PrismaClient({
 });
 
 const ROOT_DOMAIN = getRootDomain(AUTH_URL);
+
+const authLogger = createPinoOtelAdapter(createOtelLogger("hebo-auth", LOG_SEVERITY));
 
 const [
   GOOGLE_CLIENT_ID,
@@ -45,6 +49,13 @@ export const auth = betterAuth({
   baseURL: AUTH_URL,
   basePath: "/v1",
   secret: AUTH_SECRET,
+  logger: {
+    level: LOG_LEVEL as "debug" | "info" | "warn" | "error",
+    log: (level, message, ...args) => {
+      const fn = authLogger[level] ?? authLogger.info;
+      fn({ source: "better-auth", args: args.length > 0 ? args : undefined }, message);
+    },
+  },
   accountLinking: {
     enabled: true,
     trustedProviders: ["google", "github", "microsoft", "email-password"],
--- a/apps/auth/src/better-auth.ts
+++ b/apps/auth/src/better-auth.ts
@@ -4,10 +4,8 @@ import { betterAuth } from "better-auth/minimal";
 import { emailOTP, organization } from "better-auth/plugins";
 
 import { createPrismaAdapter } from "@hebo/shared-api/db/postgres";
-import { AUTH_SECRET, AUTH_URL, LOG_LEVEL, LOG_SEVERITY } from "@hebo/shared-api/env";
+import { AUTH_SECRET, AUTH_URL, LOG_LEVEL } from "@hebo/shared-api/env";
 import { COOKIE_CONFIG } from "@hebo/shared-api/lib/better-auth";
-import { createOtelLogger } from "@hebo/shared-api/lib/otel";
-import { createPinoOtelAdapter } from "@hebo/shared-api/utils/otel-pino";
 import { getSecret } from "@hebo/shared-api/utils/secret";
 import { getRootDomain } from "@hebo/shared-api/utils/url";
 
@@ -19,6 +17,7 @@ import {
   sendOrganizationInvitationEmail,
   sendVerificationOtpEmail,
 } from "./lib/email";
+import { logger } from "./lib/logger";
 import { createOrganizationHook, syncActiveOrganizationHook } from "./lib/organization";
 
 const prisma = new PrismaClient({
@@ -26,8 +25,6 @@ const prisma = new PrismaClient({
 });
 
 const ROOT_DOMAIN = getRootDomain(AUTH_URL);
-
-const authLogger = createPinoOtelAdapter(createOtelLogger("hebo-auth", LOG_SEVERITY));
 
 const [
   GOOGLE_CLIENT_ID,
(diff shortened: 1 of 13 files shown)


apps/auth/src/better-auth.ts
```diff
@@ -45,6 +49,13 @@ export const auth = betterAuth({
   baseURL: AUTH_URL,
   basePath: "/v1",
   secret: AUTH_SECRET,
+  logger: {
+    level: LOG_LEVEL as "debug" | "info" | "warn" | "error",
+    log: (level, message, ...args) => {
+      const fn = authLogger[level] ?? authLogger.info;
+      fn({ source: "better-auth", args: args.length > 0 ? args : undefined }, message);
+    },
+  },
```
heiwen: This can probably be simplified, knowing the exact types, a lot of the guards feel unnessecary.

packages/shared-api/middlewares/auth.ts
```diff
@@ -77,13 +77,17 @@ export const auth = new Elysia({ name: "auth-service" })
...
+      const { data: result, error: verifyError } = await authClient.internal.verifyApiKey({
         key: authHeader.slice(7) || "invalid-key",
         fetchOptions: {
           headers: { "x-internal-secret": AUTH_SECRET },
         },
       });
 
+      if (verifyError) {
+        logger.warn({ error: verifyError }, "API key verification request failed");
+      }
```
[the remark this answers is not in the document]
turisanapo: We are already injecting trace_id/span_id into every log. Correlation is covered.

apps/auth/src/better-auth.ts
turisanapo: Simplified.

heiwen: Why is that "source" field needed?

turisanapo: Not required, and actually redundant. Removed.

heiwen: 1/ should args still be wrapped in another object now?
2/ the log level cast should be avoidable too

turisanapo: 1/ `args` is typed `any[]`, and better-auth rarely passes extra arguments after the message. I simplified by removing it.

2/ Our `LOG_LEVEL` is `string`, better-auth's `Logger.level` expects `"debug" | "info" | "warn" | "error"`. The cast is needed to satisfy their type.

heiwen: 1/ what do you mean with "rarely"

2/ in that case our log level needs to be fixed, instead of introducing an insecure cast

heiwen: 1/ In the packages/better-auth package specifically, I found 186 logger calls in non-test source files, and 88 of them pass additional args after the main message. That’s about 47%. Most of those are error(...) calls: 81 of the 88 extra-arg cases.

2/ You just moved the cast to a different place, it remains insecure - unless you actually type guard the value. We should also have an existing log level type that you can use, no need to redefine the allowed values here.

turisanapo: 1/ You are right, my initial analysis excluded several calls where additional args are passed. I restored it.

To reply to your original question on this: our pino-style adapter accepts 2 positional arguments (`obj, message`), we need to wrap all the additional arguments into a single object.

2/ Good point. Updated.

packages/shared-api/utils/otel-pino.ts
```diff
@@ -104,11 +104,12 @@ export const createPinoOtelAdapter = (otelLogger: Logger) => {
...
 
-export const parseLogSeverity = (raw: string): SeverityNumber => {
+export const parseLogLevel = (raw: string): { level: LogLevel; severity: SeverityNumber } => {
   if (!(raw in OTEL_SEVERITY_BY_LEVEL)) {
     throw new Error(
       `Unsupported LOG_LEVEL "${raw}". Must be one of: ${Object.keys(OTEL_SEVERITY_BY_LEVEL).join(", ")}`,
     );
   }
-  return OTEL_SEVERITY_BY_LEVEL[raw as LogLevel];
+  const level = raw as LogLevel;
```
heiwen: Nitpick: inline, to be consistent with severity

apps/auth/src/better-auth.ts
```diff
@@ -50,9 +50,9 @@ export const auth = betterAuth({
   basePath: "/v1",
   secret: AUTH_SECRET,
   logger: {
-    level: LOG_LEVEL,
-    log: (level, message) => {
-      authLogger[level](message);
+    level: LOG_LEVEL === "trace" ? "debug" : LOG_LEVEL,
+    log: (level, message, ...args) => {
+      authLogger[level]({ args }, message);
```
heiwen: Open question remains: why wrap args in another object?

packages/shared-api/utils/otel-pino.ts
turisanapo: Inlined

apps/auth/src/better-auth.ts
```diff
@@ -25,6 +27,8 @@ const prisma = new PrismaClient({
 
 const ROOT_DOMAIN = getRootDomain(AUTH_URL);
 
+const authLogger = createPinoOtelAdapter(createOtelLogger("hebo-auth", LOG_SEVERITY));
```
heiwen: This creates a second adapter (which is probably fine) but also otelLogger (and with that LoggerProvider / BatchLogRecordProcessor / OTLPLogExporter) in the same project (there is already one created by the logging middleware).

Can this lead to worst case race conditions on export or at least inefficient export and should it reuse the existing one created in shared-api instead?

(We have the same problem in the gateway project)

turisanapo: I refactored the logging middleware to accept an existing logger, so that the auth and gateway services can create it once and reuse it.

heiwen: `lib/logger.ts` in `auth` feels like an unnecessary indirection, especially since it is a one-line file. just make `createLogger` return an existing logger keyed on the `serviceName` string.

heiwen: It would then be a `getLogger` (since often it doesn't create)

heiwen (approved):

PR #405, opened 2026-04-16
downgrade 4xx log severity from error to warn

### LINKED ISSUE #403: Downgrade 4xx observability severity to align with OTel semantic conventions
## Problem

All HTTP errors are currently treated equally in our observability stack — both 4xx (client errors) and 5xx (server errors) are reported as `ERROR` severity in logs and marked as `SpanStatusCode.ERROR` on OTel spans. This inflates error rates and makes dashboards/alerts noisy with client-side issues that aren't actionable by the server team.

Evidence: 8monkey-ai/hebo-cloud#5 shows inflated error values caused by this behavior.

## Proposed change

Two corrections, applied consistently across all services:

### 1. Downgrade 4xx log severity from `error` to `warn`

In `packages/shared-api/middlewares/logging.ts`, the `onError` handler currently logs everything at `logger.error`. It should split by status code:

- **4xx** → `logger.warn` (client-side issue, worth noting but not actionable by the server team)
- **5xx** → `logger.error` (server-side failure, needs investigation)

This is an industry-wide best practice across logging frameworks and observability platforms — `ERROR` should only fire for conditions the server team can act on.

### 2. Align `SpanStatusCode` to OTel HTTP semantic conventions

The [OTel HTTP server span specification](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status) states:

> Span status MUST be left **unset** if the HTTP status code was in the 1xx, 2xx, 3xx, or 4xx ranges.
> Span status MUST be set to **Error** if the HTTP status code was 5xx.

Today, all non-200 responses get `SpanStatusCode.ERROR`. The correct behavior:

- **4xx** → `UNSET` (don't call `setStatus`; the default)
- **5xx** → `SpanStatusCode.ERROR`

`span.recordException()` should continue to be called for all errors — it records diagnostic data without affecting span status.

## Affected components

The fix in `packages/shared-api/middlewares/logging.ts` (log severity) can be done directly in this repo. However, the span status fix requires changes in two additional repositories:

- **[`8monkey-ai/hebo-gateway`](https://github.com/8monkey-ai/hebo-gateway)** — `span.recordError()` in `lifecycle.js` unconditionally sets `SpanStatusCode.ERROR` for all non-200 responses. It should only do so for `>= 500`.
- **[`elysiajs/opentelemetry`](https://github.com/elysiajs/opentelemetry)** — the `inspect('Error')` trace handler sets `SpanStatusCode.ERROR` for any error that flows through Elysia's `onError`, regardless of HTTP status. It should check the resolved status code and only set `ERROR` for `>= 500`. Our fork (`turisanapo/opentelemetry`) will include the upstream PR until it is merged, at which point the fork can be retired.

If we agree on this direction, we'll open separate issues in those repos to track the span status changes.

## Impact on the traces dashboard

The `/traces` endpoint in `apps/api` determines error status from the `span_status_code` column in Greptime. After this change, 4xx errors (including invalid body 400s, model-not-found 404s, and client-caused upstream rejections) will no longer appear when filtering by `?status=error`. The `?status=error` filter will only return genuine server failures and upstream provider outages.

If we want to preserve the ability to filter by client errors in the dashboard, we can add an `http.response.status_code`-based filter as a follow-up — but this should be a deliberate product decision, not a side effect of incorrect span status.

turisanapo:
Split the onError handler so client errors (4xx) log at warn level while server errors (5xx) continue to log at error level. This reduces alert noise from non-actionable client-side issues.

Closes #403

### DIFF
--- a/packages/shared-api/middlewares/logging.ts
+++ b/packages/shared-api/middlewares/logging.ts
@@ -24,13 +24,9 @@ export const logging = (serviceName?: string) => {
   }
 
   app.onError(function logRequestError({ error, set }) {
-    logger.error(
-      {
-        status: typeof set.status === "number" ? set.status : 500,
-        err: error,
-      },
-      "request:error",
-    );
+    const status = typeof set.status === "number" ? set.status : 500;
+    const log = status >= 500 ? logger.error : logger.warn;
+    log.call(logger, { status, err: error }, "request:error");
   });
 
   return app.as("scoped");


heiwen:
I'm not sure this is correct / needed. OnError only gets real exceptions, doesn't it? All these are real 500s, aren't they?

turisanapo:
Elysia's `onError` fires for **all** thrown exceptions, not just 500s. Any `throw` in a handler or middleware triggers it.

Concrete examples that hit this handler today:

- `AuthError` (401) — thrown by `checkIsSignedIn` on every unauthenticated request
- `BadRequestError` (400) — invalid metadata filters, dual credentials in auth
- `NotFoundError` (404) — Prisma `P2025` mapped via `identifyPrismaError`
- `ConflictError` (409) — Prisma `P2002` duplicate key
- Elysia's built-in `NOT_FOUND` — any request to a non-existent route

I confirmed this from prod logs. In the last 24 hours, there are **34,380** `request:error` log records at `ERROR` severity — **100% are 4xx, zero are 5xx**:

| Service | Error type | Count |
|---|---|---|
| hebo-mcp | NOT_FOUND (404) | 16,704 |
| hebo-gateway | NOT_FOUND (404) | 16,673 |
| hebo-api | NOT_FOUND (404) | 975 |
| hebo-gateway | Unauthorized (401) | 28 |

The 28 Unauthorized entries are `AuthError` exceptions thrown by `checkIsSignedIn` — direct proof that auth errors flow through `onError` and get logged at `ERROR` severity.

This is also the root cause of the inflated error values reported in https://github.com/8monkey-ai/hebo-cloud/issues/5.

heiwen (approved):

PR #412, opened 2026-04-20
fix gateway typecheck errors against @hebo-ai/gateway@0.10.7

turisanapo:
## Summary

- **`errors.ts`**: Pass required `ResponseInit` second argument to `toErrorResponse` — the `@hebo-ai/gateway@0.10.7` API requires both `(error, init)` args, but the non-HttpError fallback path was only passing one.
- **`hooks.test.ts`**: Replace `satisfies ResolveProviderHookContext` with `as unknown as ResolveProviderHookContext` for the minimal mock context, since the mock intentionally omits many fields the full type requires.
- **`hooks.test.ts`**: Assert on `statusText` instead of `code` — `GatewayError` stores the error code in `statusText`, not a `code` property.

## Test plan

- [x] `bun run typecheck` passes with 0 errors across all workspaces
- [x] `bun test src/lib/hooks.test.ts` — all 7 tests pass

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/gateway/src/middlewares/errors.ts
+++ b/apps/gateway/src/middlewares/errors.ts
@@ -1,15 +1,8 @@
-import { GatewayError } from "@hebo-ai/gateway";
 import { toAnthropicErrorResponse } from "@hebo-ai/gateway/errors/anthropic";
 import { toOpenAIErrorResponse } from "@hebo-ai/gateway/errors/openai";
 import { Elysia } from "elysia";
 
 import { HttpError } from "@hebo/shared-api/errors";
-
-function getResponseInit(error: unknown): ResponseInit {
-  if (error instanceof HttpError) return { status: error.status };
-  if (error instanceof GatewayError) return { status: error.status, statusText: error.statusText };
-  return { status: 500 };
-}
 
 export const gatewayErrors = new Elysia({ name: "error-handler" })
   .onError(function handleGatewayError({ error, path }) {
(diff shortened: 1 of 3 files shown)


apps/gateway/src/middlewares/errors.ts
```diff
@@ -1,17 +1,22 @@
+import { GatewayError } from "@hebo-ai/gateway";
 import { toAnthropicErrorResponse } from "@hebo-ai/gateway/errors/anthropic";
 import { toOpenAIErrorResponse } from "@hebo-ai/gateway/errors/openai";
 import { Elysia } from "elysia";
 
 import { HttpError } from "@hebo/shared-api/errors";
 
+function getResponseInit(error: unknown): ResponseInit {
+  if (error instanceof HttpError) return { status: error.status };
+  if (error instanceof GatewayError) return { status: error.status, statusText: error.statusText };
```
heiwen: You shouldn't get any GatewayErrors in this handler, they should all be wrapped already. Aren't they?

apps/gateway/src/middlewares/errors.ts
```diff
@@ -1,17 +1,22 @@
+import { GatewayError } from "@hebo-ai/gateway";
 import { toAnthropicErrorResponse } from "@hebo-ai/gateway/errors/anthropic";
 import { toOpenAIErrorResponse } from "@hebo-ai/gateway/errors/openai";
 import { Elysia } from "elysia";
 
 import { HttpError } from "@hebo/shared-api/errors";
 
+function getResponseInit(error: unknown): ResponseInit {
```
heiwen: Single use function, just inline.

apps/gateway/src/middlewares/errors.ts
turisanapo: Yes, reverted.

apps/gateway/src/middlewares/errors.ts
turisanapo: Done.

heiwen (approved):

PR #415, opened 2026-04-20
fix: decorate logger in logging middleware so auth plugin receives real instance

turisanapo:
## Summary

- The refactor in `4b56c01e` ("improve auth failure observability") removed `.decorate("logger", logger)` from the `logging` middleware while adding `logger.info/warn` calls to the `auth` plugin.
- The `auth` plugin declares `.decorate("logger", {} as Logger)` as a type-level fallback, but without the real logger being decorated by `logging`, the empty `{}` was never overridden at runtime.
- In production (minified), this crashed with `B.info is not a function` on every unauthenticated request — the gateway returned a 500 instead of a proper 401.

## Fix

One-line change: re-add `.decorate("logger", logger)` to the `logging` middleware so Elysia propagates the real logger instance to downstream plugins like `auth`.

## Verified

- `bun run typecheck` passes (gateway, api, shared-api)
- `bun run -F @hebo/gateway test` — 34/34 pass
- Local `curl` with no credentials returns `401 Unauthorized` and the server correctly logs `"Authentication failed or no credentials provided"` via the real logger

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Logging now tolerates a missing logger to prevent errors when none is configured.

* **Improvements**
  * Authentication now receives logging consistently via the middleware, improving diagnostic messages.

* **Tests**
  * Added a test suite covering logging and its interaction with authentication.

* **Chores**
  * Added a package test script to run the new tests.


heiwen (approved):

heiwen:
Could have been caught before deployment. What checks do we have to introduce? How can we automate these?

heiwen:
Shouldn't the test cover whether the logger is actually working, not whether the auth is working?

turisanapo:
> Shouldn't the test cover whether the logger is actually working, not whether the auth is working?

Tests on auth were implicitly testing the logging middleware too.

Added an explicit test to check that the logging middleware decorates a working logger for downstream plugins.

heiwen:
Feels too specific. Kind of a vanity test, since it just tests that one line of code. Which is there anyway now.

A proper test whether logger works in (i) non-auth (ii) auth scenarios would be more appropriate. Also need to test that the intent of the original assignment of  "{} as Logger" is preserved.

heiwen:
Also need to revert the change in the auth middleware to decorate it's own logger. the logger should only be decorated by the logger middleware, which is then used by the auth middleware.

PR #430, opened 2026-04-21
fix: rethrow upstream auth errors instead of swallowing as 401

### LINKED ISSUE #427: Gateway should return 429 (not 401) when auth service is rate-limited
## Context

Ref #397 — When the auth service returns a 429 (Too Many Requests) on `POST /internal/verify-api-key`, the gateway swallows the error and returns 401 (Unauthorized) to the client.

This happens because:

1. `verifyApiKey` returns `{ data: null, error: { status: 429, ... } }`
2. The `resolve` step logs the error but still returns `{ userId: undefined, organizationId: undefined }`
3. The `isSignedIn` macro only checks for the presence of `userId`/`organizationId` and throws a generic `AuthError("Unauthorized")`

This is incorrect — a rate limit from the auth service is a transient infrastructure issue, not an authentication failure. Clients receiving 401 may invalidate credentials or stop retrying, when they should back off and retry.

## Proposed Fix

In the gateway auth middleware, detect non-auth errors (429, 500, 502, 503, etc.) from the `verifyApiKey` call and propagate the appropriate status code to the client instead of falling through to the 401 path.

```typescript
const { data: result, error: verifyError } = await authClient.internal.verifyApiKey({...});
if (verifyError && verifyError.status === 429) {
  throw new HTTPException(429, { message: "Too many requests" });
}
```

## Acceptance Criteria

- When auth service returns 429, the gateway returns 429 to the client (not 401)
- When auth service returns 5xx, the gateway returns 502 or 503 (not 401)
- Normal auth failures (invalid key, expired key) still return 401
- Clients see correct `Retry-After` header if provided by the auth service

Relates to #397

heiwen:
When `verifyApiKey` returns any error (429, 5xx, etc.), throw an `HttpError` with the original status code instead of falling through to the 401 path.

Closes #427

## Summary by CodeRabbit

* **Bug Fixes**
  * API key verification failures now properly return HTTP error responses to clients instead of being silently logged.

### DIFF
--- a/packages/shared-api/middlewares/auth.ts
+++ b/packages/shared-api/middlewares/auth.ts
@@ -87,7 +87,6 @@ export const auth = new Elysia({ name: "auth-service" })
       });
 
       if (verifyError) {
-        logger.warn({ error: verifyError }, "API key verification request failed");
         throw new HttpError(verifyError.message ?? "Auth service error", verifyError.status ?? 500, "upstream_auth_error");
       }
 
--- a/packages/shared-api/middlewares/auth.ts
+++ b/packages/shared-api/middlewares/auth.ts
@@ -7,7 +7,7 @@ import { type Cookie, Elysia } from "elysia";
 import type { VerifyApiKeyPlugin } from "~auth/lib/api-key";
 
 import { AUTH_SECRET, AUTH_URL } from "../env";
-import { AuthError, BadRequestError } from "../errors";
+import { AuthError, BadRequestError, HttpError } from "../errors";
 import { COOKIE_CONFIG } from "../lib/better-auth";
 import { logging } from "./logging";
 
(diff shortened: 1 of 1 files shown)


packages/shared-api/middlewares/auth.ts
```diff
@@ -88,6 +88,7 @@ export const auth = new Elysia({ name: "auth-service" })
 
       if (verifyError) {
         logger.warn({ error: verifyError }, "API key verification request failed");
```
turisanapo: Do we still need the explicit `logger.warn()`, considering the error will be logged automatically by the `logging` plugin?

heiwen:
@claude what do you think about @turisanapo comment? is he right that there is already an existing log line?

heiwen:
@claude I'm ok with removing logger.warn then

packages/shared-api/middlewares/auth.ts
heiwen: Removed

turisanapo (approved):

PR #464, opened 2026-05-11
fix(console): exclude toggle labels from trace message copy

### LINKED ISSUE #460: Chat history message copy includes 'less'/'more' toggle labels
## Description

In the console, the `/traces` section displays the chat history. When a message is too long, a "less"/"more" toggle is shown to expand/collapse the message block. When using the copy button on a message block, the "less" or "more" label text is included in the clipboard content.

## Steps to Reproduce

1. Navigate to the `/traces` section in the console
2. Find a message that is long enough to trigger the "less"/"more" toggle
3. Click the copy button on the message block
4. Paste the copied content

## Expected Behaviour

The full content of the message is copied without the "less" or "more" labels.

## Observed Behaviour

The copied content includes the "less" or "more" label text depending on the current expand/collapse state of the message block.

heiwen:
## Summary
- Teach `CopyButton` to accept a DOM subtree via its `value` prop (now `string | () => string | RefObject<HTMLElement>`) and walk the tree itself, skipping any subtree marked with `data-copy-ignore`.
- Annotate the trace message toggles, the Reasoning disclosure trigger, and the tool-call / other-part label elements with `data-copy-ignore`.
- Drop `getNodeText` from `Code.tsx` and point `CopyButton` at the `<pre>` ref so there's one copy of the text in the DOM and the clipboard string is only built on click.
- Walker uses `textContent` semantics so collapsed / hidden text is still copied in full (no `line-clamp` truncation, Reasoning panel uses `keepMounted`).

Fixes #460

## Test plan
- [ ] Long message still collapsed: copy → full text, no "More" / "Less".
- [ ] Long message expanded: copy → full text, no "More" / "Less".
- [ ] Message with reasoning (collapsed): copy → reasoning text is present, "Reasoning" trigger label is not.
- [ ] Message with tool calls: copy → arguments present, tool-call badge name absent.
- [ ] CodeBlock copy → pasted code matches rendered content (title and untitled variants).
- [ ] Other `CopyButton` sites (Raw JSON, metadata rows, alias paths, Input) still copy expected string.

## Summary by CodeRabbit

* **Improvements**
  * Copy-to-clipboard now ignores UI controls (badges, expand/collapse triggers) so copied content is cleaner
  * Copy behavior expanded to derive text from rendered elements for more accurate results when copying code or message blocks
  * Expandable content now keeps children mounted after collapse to preserve state and improve responsiveness

[Review Change Stack](https://app.coderabbit.ai/change-stack/8monkey-ai/hebo-platform/pull/464)

### DIFF
--- a/apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.traces.($traceId)/details.tsx
+++ b/apps/console/app/routes/_shell.agent.$agentSlug.branch.$branchSlug.traces.($traceId)/details.tsx
@@ -341,7 +341,7 @@ function MessageBlock({ message }: { message: TraceMessage }) {
             )}
           </div>
 
-          <CopyButton value={() => contentRef.current?.innerText ?? ""} />
+          <CopyButton sourceRef={contentRef} />
         </div>
 
         {!reasoning && texts.length === 0 && toolCalls.length === 0 && otherParts.length === 0 ? (
@@ -369,7 +369,7 @@ function MessageBlock({ message }: { message: TraceMessage }) {
 
             {toolCalls.map((tc, index) => (
               <div key={`tool-call:${index}`} className="space-y-2">
-                <Badge variant="outline">
+                <Badge variant="outline" data-copy-ignore>
                   <Wrench className="size-3" />
                   {tc.name}
                 </Badge>
@@ -379,7 +379,10 @@ function MessageBlock({ message }: { message: TraceMessage }) {
 
             {otherParts.map((part, index) => (
               <div key={`${part.type}:${index}`} className="space-y-2">
-                <div className="text-xs font-medium text-muted-foreground uppercase">
+                <div
+                  data-copy-ignore
+                  className="text-xs font-medium text-muted-foreground uppercase"
+                >
                   {part.type}
                 </div>
                 <CollapsibleCode code={part.value} maxLength={300} />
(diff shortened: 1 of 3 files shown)


heiwen:
@claude what do you think about the codereview comments?

heiwen:
@claude how was the whitespace normalization handled before?

heiwen:
@claude why do we need the BLOCK_TAG_PATTERN in the first place, can't we simplify that?

heiwen:
@claude it feels more complicated than it needs to be. why not simply walk the whole HTML tree recursively, like it was done previously in getNodeText()?

heiwen:
@claude ok, go with this version:

```
const BLOCK_TAGS = new Set(["DIV", "P", "PRE"]);

function collectCopyText(root: HTMLElement): string {
  if (root.tagName === "PRE") return root.textContent ?? "";

  const walk = (node: Node): string => {
    if (node.nodeType === Node.TEXT_NODE) return node.nodeValue ?? "";
    if (node.nodeType !== Node.ELEMENT_NODE) return "";
    const el = node as HTMLElement;
    if (el.hasAttribute(COPY_IGNORE_ATTR)) return "";
    if (el.tagName === "BR") return "\n";
    const inner = Array.from(el.childNodes).map(walk).join("");
    return BLOCK_TAGS.has(el.tagName) ? `\n${inner}\n` : inner;
  };

  return walk(root).replaceAll(/\n+/g, "\n").trim();
}
```

heiwen:
@claude ok, so we are using the DOM representation now. Would the alternative be to use the virtual dom and if so how to do that without creating duplicate representations of everything just for the copy button?

heiwen:
@claude I won't the CopyButton to be generic, so that the client doesn't have to provide their own mapping logic every single time. but I also like to avoid the keepMounted, since that can slow down the frontend significantly. how?

heiwen:
@claude there must be another way. my constraints are clear (i) keep using the collapsible component, we don't have access to it's API (ii) don;t use keepMounted (iii) a generic way to pass the content into the copybutton [whether that's dom or virtual dom or whatever I don't know). (iv) no expensive duplication so it doesn't blow up memory or the like. this should be possible, since the content is already there somewhere. What's that way? (don't implement anything)

PR #472, opened 2026-06-03
chore: upgrade oxlint to 1.68 and oxlint-tsgolint to 0.22.1

turisanapo:
## Summary

- Upgrades `oxlint` from `^1.60.0` to `^1.68.0` and `oxlint-tsgolint` from `^0.21.1` to `^0.22.1`
- Fixes tsgolint SIGPIPE crashes caused by peer dependency mismatch (oxlint 1.68 requires tsgolint >=0.22.1)
- Resolves newly enforced lint rules from the `pedantic` category added in oxlint 1.61–1.68

## Changes

| Rule | Fix |
|------|-----|
| `require-unicode-regexp` | Added `u` flag to all regex literals |
| `no-underscore-dangle` | Suppressed for HMR global (`__heboDb`), renamed `_fetch` → `originalFetch` |
| `no-object-type-as-default-prop` | Extracted JSX default values to module-level constants |
| `prefer-tag-over-role` | Suppressed for custom components that forward `role` to internal DOM elements |

## Test plan

- [x] `bun run typecheck` passes (exit code 0)
- [x] `bun run lint` passes (exit code 0)

## Summary by CodeRabbit

* **Chores**
  * Bumped developer tooling versions (`oxlint`, `oxlint-tsgolint`)
  * Minor runtime/init tweaks (global HMR variable rename, fetch wrapper override)

* **Refactor**
  * Improved pattern-matching by enabling Unicode-aware regexes across multiple modules
  * Extracted default component constants for icons

* **Style**
  * Added ESLint suppression comments for accessibility-related roles

### DIFF
--- a/apps/console/app/mocks/db.ts
+++ b/apps/console/app/mocks/db.ts
@@ -65,7 +65,9 @@ export const createDb = () => {
 type DB = ReturnType<typeof createDb>;
 
 declare global {
+  // oxlint-disable-next-line no-underscore-dangle -- HMR global singleton pattern
   var __heboDb: DB | undefined;
 }
 
+// oxlint-disable-next-line no-underscore-dangle -- HMR global singleton pattern
 export const db: DB = (globalThis.__heboDb ??= createDb());
--- a/apps/console/app/mocks/db.ts
+++ b/apps/console/app/mocks/db.ts
@@ -65,9 +65,7 @@ export const createDb = () => {
 type DB = ReturnType<typeof createDb>;
 
 declare global {
-  // oxlint-disable-next-line no-underscore-dangle -- HMR global singleton pattern
-  var __heboDb: DB | undefined;
+  var heboDb: DB | undefined;
 }
 
-// oxlint-disable-next-line no-underscore-dangle -- HMR global singleton pattern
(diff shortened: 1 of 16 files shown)


apps/console/app/mocks/db.ts
```diff
@@ -65,7 +65,9 @@ export const createDb = () => {
 type DB = ReturnType<typeof createDb>;
 
 declare global {
+  // oxlint-disable-next-line no-underscore-dangle -- HMR global singleton pattern
```
heiwen: Feels weird to disable a rule in a mock. Do we consider this rule as generally helpful? If so, why wouldn't be address it here? If not, should we globally disabled.

turisanapo: I find the rule useful, it discourages using `_foo`/`__foo` as a pseudo-private convention, which is just a convention and not enforced by the runtime. The correct way would be to use `#field` when defining the class field. However, it can't be used on `globalThis`, and this is why I initially disabled the rule.

Anyway, since no other code in the repo other than the mock would use `heboDb`, I renamed it to avoid explicit suppression.

heiwen:
LGTM

PR #480, opened 2026-07-31
feat(docker): inject console URLs as build args for self-hosted deploys

### LINKED ISSUE #477: Console use localhost instead of actual domain in self hosted deploy
### Context

Hebo console in standalone deployment requires `VITE_API_URL/VITE_AUTH_URL/VITE_GATEWAY_URL` which fallback to localhost.

This works in local, but in QA or production, URL are custom domains it requires us to inject them via env variables.

Current Dockerfile in self hosted standalone does not inject those variable at build time and fallback to localhost which is unexpected.

### Suggestion

Option 1: Inject those variables via Dockerfile to build docker image with expected domains.
- baking URLs makes image environment specific
- anyone has their own domain has to rebuild it

Option 2: Make domain configs runtime injection instead of build time
- this introduces some complexity, caching ability 
- any misconfiguration won't be caught at build time

Option 3: Serve console with same-origin with /api/v1, /auth/v1, /gateway/v1 path prefixes in standalone
- this requires bigger change scope that feels overkill

_Given it is simpler and safer, option 1 is recommended_

buibaoanh:
Closes #477

## Problem

The console is a static SPA, so `VITE_API_URL` / `VITE_AUTH_URL` / `VITE_GATEWAY_URL` are compiled into the bundle by Vite at build time (`apps/console/app/lib/env.ts`). `infra/docker/Dockerfile` passed none of them, so the published image always fell back to `localhost:852x` and self-hosted deployments behind a custom domain pointed the console at the user's own machine.

## Approach

Option 1 from the issue — bake the URLs at build time.

- `infra/docker/Dockerfile` accepts the four `VITE_*` values as build args, declared **after** `bun install` so changing a domain reuses the dependency layer instead of reinstalling.
- Unset args stay empty and `env.ts` keeps its `||` fallback, so the published image is byte-for-byte equivalent in behaviour to today.
- `infra/self-hosted/docker-compose.build.yaml` (new) is an opt-in override that builds locally with those args from `.env`. Kept separate so the default `docker compose up -d` still pulls the prebuilt image and the README quickstart doesn't silently become a from-source build.
- `HEBO_IMAGE` makes the image tag overridable, so a locally built image needn't clobber the official `:latest` tag.

## Verification

Two real image builds of the `builder` stage:

| Build | `api/auth/gateway.example.com` in bundle | `localhost:8521/2/3` in bundle |
| --- | --- | --- |
| With build args | 1 / 1 / 1 | 0 / 0 / 0 |
| No build args (published-image path) | — | 1 / 1 / 1 |

URLs land in `build/client/assets/env-*.js`; Vite dead-code-eliminates the localhost fallbacks when a value is supplied. Merged compose config verified to resolve the build context to the repo root, interpolate all four args from `.env`, and apply the `HEBO_IMAGE` override.

## Note on scope

Baking URLs means anyone on a custom domain must build their own image rather than pull `8monkey/hebo-platform` — a known tradeoff of Option 1, called out in the issue and accepted.

Setting the `VITE_*` values alone is also not sufficient for a working custom-domain deployment: `BASE_URL` drives CORS, the better-auth base URL, trusted origins and the session cookie domain, and the self-hosted compose file sets none of it. This PR **documents** `BASE_URL` / `AUTH_URL` / `NODE_ENV` in the custom-domain instructions but changes no server-side behaviour. Worth separate issues:

- With `BASE_URL` unset, `packages/shared-api/lib/cors.ts:5` accepts **every** origin with `credentials: true`, and `apps/auth/src/better-auth.ts:88` sets `trustedOrigins: ["*"]` — wide open by default for a publicly exposed self-hosted deploy.
- `trustedOrigins` is `https://*.${ROOT_DOMAIN}` (subdomains only) while CORS also allows the apex, so a console on the apex domain passes CORS then fails `callbackURL` validation.
- `apps/mcp/src/index.ts` has no CORS and no auth middleware, and compose exposes port 8524.
- `VITE_MAGICLINK_AUTH` is baked while the auth server gates `emailAndPassword` on SMTP at runtime, so the two can disagree.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for configuring custom domains, HTTPS, OAuth callbacks, cookies, CORS, and OpenAPI URLs in self-hosted deployments.
  * Documented frontend build-time configuration and deployment environment variables.

* **New Features**
  * Added support for building custom self-hosted images with configurable API, authentication, gateway, and magic-link URLs.
  * Made the self-hosted image configurable, with the standard image used by default.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -201,6 +201,8 @@ The Docker image is built from `infra/docker/Dockerfile` and published to Docker
 - **Release** — pushing a `v*` tag builds with `NODE_ENV=production` and tags the image as `v<semver>`
 - **Manual** — trigger the workflow manually for development builds
 
+The console's `VITE_API_URL`, `VITE_AUTH_URL`, `VITE_GATEWAY_URL` and `VITE_MAGICLINK_AUTH` are Vite constants baked into the static bundle, so the Dockerfile takes them as build args. The published image is built without them and falls back to `localhost`; self-hosters on a custom domain build their own image via `infra/self-hosted/docker-compose.build.yaml`.
+
 ### Cloud (AWS ECS)
 
 The repository uses GitHub Actions for CI/CD with SST managing the infrastructure:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -201,7 +201,7 @@ The Docker image is built from `infra/docker/Dockerfile` and published to Docker
 - **Release** — pushing a `v*` tag builds with `NODE_ENV=production` and tags the image as `v<semver>`
 - **Manual** — trigger the workflow manually for development builds
 
-The console's `VITE_API_URL`, `VITE_AUTH_URL`, `VITE_GATEWAY_URL` and `VITE_MAGICLINK_AUTH` are Vite constants baked into the static bundle, so the Dockerfile takes them as build args. The published image is built without them and falls back to `localhost`; self-hosters on a custom domain build their own image via `infra/self-hosted/docker-compose.build.yaml`.
+The console's `VITE_*` values are Vite constants baked into the bundle, so the Dockerfile takes them as build args. The published image is built without them and falls back to `localhost`; custom-domain deployments build their own image via `infra/self-hosted/docker-compose.build.yaml`.
 
 ### Cloud (AWS ECS)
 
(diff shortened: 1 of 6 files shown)


buibaoanh:
Close this PR in favor of https://github.com/8monkey-ai/hebo-platform/pull/485

PR #485, opened 2026-08-12
feat(console): resolve service URLs at container start, not build time

### LINKED ISSUE #477: Console use localhost instead of actual domain in self hosted deploy

buibaoanh:
Closes #477. Alternative to #480 — same problem, runtime config instead of build args.

## Problem

The console is a static SPA, so `VITE_API_URL` / `VITE_AUTH_URL` / `VITE_GATEWAY_URL` / `VITE_MAGICLINK_AUTH` are compiled into the bundle by Vite (`apps/console/app/lib/env.ts`). The published image always falls back to `localhost:852x`, so a self-hosted deployment behind a custom domain points the user's browser at their own machine.

## Approach

`#480` bakes the URLs in as build args, which means anyone on a custom domain has to build their own image. This resolves them at container start instead, so the published image works as-is.

`apps/console/serve.ts` already serves `index.html`, so it now inlines its own environment as `window.heboEnv` at the top of `<head>`. A classic inline script runs before the deferred module bundle, so `env.ts` sees the config on first read and lets it override whatever Vite baked in.

Deliberately unchanged:

- **No Dockerfile changes.** Nothing is needed at build time.
- **No compose changes.** `docker-compose.yaml` already forwards `.env` via `env_file`, and the console's s6 service already uses `with-contenv` — the same path `api` uses for `POSTGRES_URL`.
- **`infra/stacks/console.ts` (cloud) untouched.** SST `StaticSite` has no server to inject from, so it keeps using the build-time values, which are still honoured.
- **Kept the `VITE_` names** so the cloud build-time path and the self-hosted runtime path share one set of variable names. `VITE_` is a slight misnomer at runtime; two names would be worse.

Unset keys are omitted from the injected object, so the build-time value and the `localhost` fallbacks still apply — the published image behaves exactly as it does today.

Index HTML is served `cache-control: no-cache` since the document now carries the config; a stale copy would point at the wrong domain. Values are `<`-escaped so one containing `</script>` can't break out of the tag, and a missing `<head>` throws at startup rather than silently dropping the config.

## Verification

`bun test` (4 new tests in `apps/console/serve.test.ts`), `bun run lint` and `bun run typecheck` all clean.

Built the console with **no** `VITE_*` set — the published-image path — then checked the shipped `build/client/assets/env-*.js`:

| | `apiUrl` / `authUrl` / `gatewayUrl` |
| --- | --- |
| No runtime env | `localhost:8521` / `8523` / `8522` — identical to today |
| `heboEnv` injected | `https://api.example.com` / `auth.example.com` / `gw.example.com` |

End-to-end in headless Chromium against `bun serve.ts` with the vars set: the page sees `window.heboEnv` with all three values, and the app's env module resolves to the injected URLs. SPA fallback still returns 200, static assets 200, missing assets 404.

## Note on scope

Setting the console URLs alone isn't enough for a working custom-domain deployment — `BASE_URL` drives CORS, the better-auth base URL, trusted origins and the session cookie domain, and the self-hosted compose file sets none of it. This PR **documents** `BASE_URL` in the new custom-domain section and `.env.example` but changes no server-side behaviour. Still worth separate issues (all carried over from #480, none addressed here):

- With `BASE_URL` unset, `packages/shared-api/lib/cors.ts:5` accepts **every** origin with `credentials: true`, and `apps/auth/src/better-auth.ts:88` sets `trustedOrigins: ["*"]` — wide open by default for a publicly exposed deploy.
- `trustedOrigins` is `https://*.${ROOT_DOMAIN}` (subdomains only) while CORS also allows the apex, so a console on the apex domain passes CORS then fails `callbackURL` validation.
- `apps/mcp/src/index.ts` has no CORS and no auth middleware, and compose exposes port 8524.
- `VITE_MAGICLINK_AUTH` is now runtime-settable but the auth server gates `emailAndPassword` on SMTP independently, so the two can still disagree.

## Summary by CodeRabbit

* **New Features**
  * Added runtime configuration for console API, authentication, gateway, and magic-link endpoints.
  * Improved deployed console support with runtime environment injection and SPA route handling.

* **Documentation**
  * Added self-hosted custom-domain guidance covering CORS, authentication, session cookies, subdomain deployments, SMTP, and OAuth sign-in options.

* **Tests**
  * Added coverage for runtime configuration injection, including handling of empty, undefined, and non-applicable values.

### DIFF
--- a/apps/console/app/lib/env.ts
+++ b/apps/console/app/lib/env.ts
@@ -8,16 +8,38 @@ const isReachable = (url: string) =>
 
 export const useMocks = shouldAutoDetect && !(await isReachable("http://localhost:8521"));
 
+type ConsoleEnv = {
+  VITE_API_URL?: string;
+  VITE_AUTH_URL?: string;
+  VITE_GATEWAY_URL?: string;
+  VITE_MAGICLINK_AUTH?: string;
+};
+
+declare global {
+  var heboEnv: ConsoleEnv | undefined;
+}
+
+// Runtime config, injected into index.html by serve.ts, wins over the values Vite
+// baked in — that's what lets the published image be pointed at any domain without
+// a rebuild. Unset keys are omitted, so the build-time value stays in play.
+const env: ConsoleEnv = {
+  VITE_API_URL: import.meta.env.VITE_API_URL,
+  VITE_AUTH_URL: import.meta.env.VITE_AUTH_URL,
+  VITE_GATEWAY_URL: import.meta.env.VITE_GATEWAY_URL,
+  VITE_MAGICLINK_AUTH: import.meta.env.VITE_MAGICLINK_AUTH,
+  ...globalThis.heboEnv,
+};
+
 // oxlint-disable prefer-nullish-coalescing -- empty string should use the fallback URL
 export const apiUrl = useMocks
   ? "http://localhost:8520/api"
-  : import.meta.env.VITE_API_URL || "http://localhost:8521";
+  : env.VITE_API_URL || "http://localhost:8521";
 
-export const authUrl = import.meta.env.VITE_AUTH_URL || "http://localhost:8523";
+export const authUrl = env.VITE_AUTH_URL || "http://localhost:8523";
 
 export const gatewayUrl = useMocks
   ? "http://localhost:8520/gateway"
-  : import.meta.env.VITE_GATEWAY_URL || "http://localhost:8522";
+  : env.VITE_GATEWAY_URL || "http://localhost:8522";
 // oxlint-enable prefer-nullish-coalescing
 
-export const magicLinkAuth = import.meta.env.VITE_MAGICLINK_AUTH === "true";
+export const magicLinkAuth = env.VITE_MAGICLINK_AUTH === "true";
--- a/apps/console/app/lib/env.ts
+++ b/apps/console/app/lib/env.ts
@@ -8,21 +8,12 @@ const isReachable = (url: string) =>
 
 export const useMocks = shouldAutoDetect && !(await isReachable("http://localhost:8521"));
 
-type ConsoleEnv = {
-  VITE_API_URL?: string;
-  VITE_AUTH_URL?: string;
-  VITE_GATEWAY_URL?: string;
-  VITE_MAGICLINK_AUTH?: string;
-};
-
 declare global {
-  var heboEnv: ConsoleEnv | undefined;
+  var heboEnv: Record<string, string | undefined> | undefined;
 }
 
-// Runtime config, injected into index.html by serve.ts, wins over the values Vite
-// baked in — that's what lets the published image be pointed at any domain without
-// a rebuild. Unset keys are omitted, so the build-time value stays in play.
-const env: ConsoleEnv = { ...import.meta.env, ...globalThis.heboEnv };
+// Runtime config from serve.ts wins; unset keys fall through to the build-time values.
+const env = { ...import.meta.env, ...globalThis.heboEnv };
 
 // oxlint-disable prefer-nullish-coalescing -- empty string should use the fallback URL
 export const apiUrl = useMocks
--- a/apps/console/app/lib/env.ts
+++ b/apps/console/app/lib/env.ts
@@ -22,13 +22,7 @@ declare global {
 // Runtime config, injected into index.html by serve.ts, wins over the values Vite
 // baked in — that's what lets the published image be pointed at any domain without
 // a rebuild. Unset keys are omitted, so the build-time value stays in play.
-const env: ConsoleEnv = {
-  VITE_API_URL: import.meta.env.VITE_API_URL,
-  VITE_AUTH_URL: import.meta.env.VITE_AUTH_URL,
-  VITE_GATEWAY_URL: import.meta.env.VITE_GATEWAY_URL,
-  VITE_MAGICLINK_AUTH: import.meta.env.VITE_MAGICLINK_AUTH,
-  ...globalThis.heboEnv,
-};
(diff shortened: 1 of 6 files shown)


apps/console/app/lib/env.ts
```diff
@@ -8,16 +8,32 @@ const isReachable = (url: string) =>
 
 export const useMocks = shouldAutoDetect && !(await isReachable("http://localhost:8521"));
 
+type ConsoleEnv = {
```
turisanapo: Why introduce `ConsoleEnv` when we already have `import.meta.env`?

apps/console/serve.ts
```diff
@@ -1,23 +1,53 @@
...
-  },
-});
+// Public config the console resolves at runtime rather than at build time, so one
+// prebuilt image can be pointed at any domain. Consumed by app/lib/env.ts.
+const RUNTIME_ENV_KEYS = [
+  "VITE_API_URL",
+  "VITE_AUTH_URL",
+  "VITE_GATEWAY_URL",
+  "VITE_MAGICLINK_AUTH",
+] as const;
```
turisanapo: You should be able to filter for `VITE_` vars rather than re-defining the list.

apps/console/serve.ts
```diff
@@ -1,23 +1,53 @@
...
+
+/**
+ * Inlines `window.heboEnv` at the top of `<head>`. A classic inline script
+ * runs before the deferred module bundle, so the app sees the config on first read.
+ */
+export const injectRuntimeEnv = (html: string, env: Record<string, string | undefined>) => {
+  if (!html.includes("<head>")) throw new Error("serve.ts: no <head> in index.html");
```
turisanapo: Is this case possible? We control the build step that produces the HTML.

apps/console/serve.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  it("escapes < so a value cannot close the script tag", () => {
+    const out = injectRuntimeEnv(html, { VITE_API_URL: "</script><script>alert(1)</script>" });
+
+    expect(out).not.toContain("</script><script>alert(1)");
+    expect(out.match(/<script>/gu)).toHaveLength(1);
+  });
+
+  it("fails loudly rather than silently dropping config", () => {
+    expect(() => injectRuntimeEnv("<html><body></body></html>", {})).toThrow(/no <head>/u);
+  });
```
turisanapo: This shouldn't be necessary. See comments on apps/console/serve.ts

apps/console/app/lib/env.ts
```diff
@@ -8,16 +8,32 @@ const isReachable = (url: string) =>
...
+  VITE_MAGICLINK_AUTH?: string;
+};
+
+declare global {
+  var heboEnv: ConsoleEnv | undefined;
+}
+
+// Runtime config, injected into index.html by serve.ts, wins over the values Vite
+// baked in — that's what lets the published image be pointed at any domain without
+// a rebuild. Unset keys are omitted, so the build-time value stays in play.
```
turisanapo: Could you simplify this comment?

turisanapo (changes requested):

apps/console/serve.ts
buibaoanh: I added comment on why we need this config to avoid redundant injections in HTML.

turisanapo: Every `VITE_` var is already public. I'd still filter on the prefix.

apps/console/serve.ts
```diff
@@ -1,23 +1,48 @@
...
-    return new Response(indexFile);
-  },
-});
+// Public config the console resolves at runtime, so one prebuilt image can be pointed at
+// any domain. Allowlisted, not VITE_-prefix-filtered: these land in publicly served HTML.
+const RUNTIME_ENV_KEYS = [
+  "VITE_API_URL",
+  "VITE_AUTH_URL",
+  "VITE_GATEWAY_URL",
+  "VITE_MAGICLINK_AUTH",
```
turisanapo: You are whitelisting `VITE_MAGICLINK_AUTH` but not documenting it.

apps/console/serve.ts
```diff
@@ -1,23 +1,48 @@
...
+] as const;
+
+/** Inlines `window.heboEnv` into `<head>`, where it runs before the deferred module bundle. */
+export const injectRuntimeEnv = (html: string, env: Record<string, string | undefined>) => {
+  const values = Object.fromEntries(
+    RUNTIME_ENV_KEYS.filter((key) => env[key]).map((key) => [key, env[key]]),
+  );
+  // Escape `<` so a stray "</script>" in a value can't close the tag early.
+  const json = JSON.stringify(values).replaceAll("<", "\\u003c");
```
turisanapo: Is this case actually possible considering the nature of the values we expect the operator to provide? I would rather remove the guard.

turisanapo (approved):

